Issue #13, opened 2023-08-24
Current Footer in Blog Home uses 'footer-copy'  pattern instead of Footer template part


jordesign:
It appears that the theme is currently using a pattern for the footer instead of Footer Template Part which is included.

<img width="782" alt="Screenshot 2023-08-24 at 8 33 21 pm" src="https://github.com/WordPress/twentytwentyfour/assets/1842363/cb699289-788f-44d0-bb9f-df5c04df9a11">

jordesign:
PR at https://github.com/WordPress/twentytwentyfour/pull/14

PR #14, opened 2023-08-24
Make home.html use correct footer

jordesign:
Addresses https://github.com/WordPress/twentytwentyfour/issues/13

### DIFF
--- a/templates/home.html
+++ b/templates/home.html
@@ -5,5 +5,5 @@
 
 <!-- wp:pattern {"slug":"twentytwentyfour/home"} /-->
 
-<!-- wp:template-part {"slug":"footer-copy"} /--></div>
+<!-- wp:template-part {"slug":"footer","tagName":"footer"} /--></div>
 <!-- /wp:group -->


templates/home.html
```diff
@@ -5,5 +5,5 @@
 
 <!-- wp:pattern {"slug":"twentytwentyfour/home"} /-->
 
-<!-- wp:template-part {"slug":"footer-copy"} /--></div>
```
MaggieCabrera, line 8: footer-copy and footer don't look the same, we need to address having two in a different way

jordesign: Should footer-copy be a variant footer template part?

MaggieCabrera: I think there needs to be some general refactoring of the templates so we can abstract the differences into the templates instead of into the template parts.

beafialho:
`footer-copy` can be removed as it seems to be there by mistake. `footer` is the correct one and it should be called **Footer (Default)**. We'll need to make those paragraphs actual links, or use a navigation block though.

Captura de ecrã 2023-08-30, às 15 26 14

Also two more footers were designed to be alternatives, as specified in the [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=298-1644&mode=design&t=tQqcX6ylmUY0sScX-0) file:

**Footer (Writing)**
Captura de ecrã 2023-08-30, às 15 28 18

and **Footer (Portfolio)**

Captura de ecrã 2023-08-30, às 15 28 27

carolinan:
From these discussions, I think the PR can be merged? It corrects the default footer and it is a very small change if we need to replace the part later.

carolinan (approved):

PR #21, opened 2023-08-24
Update readme file

kafleg:
Removed changelog and changed copyright


jeffikus (approved):
LGTM! Makes sense to change it from 6.3 to 6.4 for when it will ship. I'll change the style sheet now.

PR #22, opened 2023-08-24
About pattern

amedv:
Generated in WordPress block editor based on [Figma template](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=492-4916&mode=dev) and added WordPress formation

### DIFF
--- a/patterns/about.php
+++ b/patterns/about.php
@@ -7,8 +7,8 @@
 
 ?>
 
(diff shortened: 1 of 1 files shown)


carolinan (approved):
This is a good start, thank you

PR #23, opened 2023-08-24
Add tagNname main for skip links

kafleg:
Skip link was not working in home page and this PR will add it.

### DIFF
--- a/patterns/home.php
+++ b/patterns/home.php
@@ -6,7 +6,7 @@
  * Inserter: no
  */
 ?>
-<!-- wp:group {"style":{"spacing":{"padding":{"top":"6vh","bottom":"6vh"}}},"layout":{"type":"constrained"}} -->
+<!-- wp:group {"tagName":"main","style":{"spacing":{"padding":{"top":"6vh","bottom":"6vh"}}},"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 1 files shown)


patterns/home.php
```diff
@@ -6,7 +6,7 @@
  * Inserter: no
  */
 ?>
-<!-- wp:group {"style":{"spacing":{"padding":{"top":"6vh","bottom":"6vh"}}},"layout":{"type":"constrained"}} -->
+<!-- wp:group {"tagName":"main","style":{"spacing":{"padding":{"top":"6vh","bottom":"6vh"}}},"layout":{"type":"constrained"}} -->
 <div class="wp-block-group" style="padding-top:6vh;padding-bottom:6vh"><!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|40","margin":{"top":"0","bottom":"0"},"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|80"}}},"layout":{"type":"flex","orientation":"vertical","justifyContent":"center"}} -->
```
MaggieCabrera, line 10: I think we need to change the html tag too, not just the comment

carolinan: Yes we do

carolinan: However, if this pattern can be placed anywhere, it should not use the `<main>` since there should only be one main per _page._

carolinan:
I see the `inserter:no` now so my comment was incorrect. This issue has already been fixed.

PR #36, opened 2023-08-24
adds a basic testimonial pattern

### LINKED ISSUE #112: Block Patterns - Testimonial
**Description**

🛠️ The goal of this issue is to track the implementation of the Testimonial pattern.
🎨 [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=492%3A4302&mode=design&t=LHSPHKUrPq556l59-1)

Testimonial

aplauche:
Current pattern uses wide alignment, could swap to default to normal width if that makes more sense.

### DIFF
--- a/patterns/testimonial.php
+++ b/patterns/testimonial.php
@@ -8,8 +8,8 @@
 ?>
 
 <!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"140px","bottom":"140px"}}},"backgroundColor":"contrast","textColor":"tertiary","layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignwide has-tertiary-color has-contrast-background-color has-text-color has-background" style="padding-top:140px;padding-bottom:140px"><!-- wp:paragraph {"align":"center","style":{"spacing":{"margin":{"bottom":"32px"}},"typography":{"lineHeight":"1.2"}},"textColor":"tertiary","fontSize":"large","fontFamily":"cardo"} -->
-  <p class="has-text-align-center has-tertiary-color has-text-color has-cardo-font-family has-large-font-size" style="margin-bottom:32px;line-height:1.2"><em><?php echo esc_html__('“Études has saved us thousands of hours of work and has unlocked insights we never thought possible.”', 'twentytwentyfour'); ?></em></p>
+<div class="wp-block-group alignwide has-tertiary-color has-contrast-background-color has-text-color has-background" style="padding-top:140px;padding-bottom:140px"><!-- wp:paragraph {"align":"center","style":{"spacing":{"margin":{"bottom":"32px"}},"typography":{"lineHeight":"1.2"}},"textColor":"tertiary","fontSize":"x-large","fontFamily":"cardo"} -->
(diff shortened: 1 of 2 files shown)


kishanjasani:
@carolinan Weather it is dark mode or light mode in styles variation. This patterns background color is Black and foreground is white. Check this link: https://prnt.sc/pOPADkpmakFO

This pattern uses `.has-tertiary-color` class [Check this code.](https://github.com/WordPress/twentytwentyfour/pull/36/files#diff-bb21203ac90c451af02b0e7cf83642398217283bc83799f195ec020c0090db27R11) which will convert background color to white and text color to black in `Dark Serif  style variation`. What could be the best solution for this. Introduce a new color variable in theme.json?

carolinan:
I recommend keeping it as is without adding more colors.

patterns/testimonial.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Title: Testimonial
+ * Slug: twentytwentyfour/testimonial
+ * Categories: testimonials
+ */
+?>
+
+<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"140px","bottom":"140px"}}},"backgroundColor":"contrast","textColor":"tertiary","layout":{"type":"default"}} -->
+<div class="wp-block-group alignwide has-tertiary-color has-contrast-background-color has-text-color has-background" style="padding-top:140px;padding-bottom:140px"><!-- wp:paragraph {"align":"center","style":{"spacing":{"margin":{"bottom":"32px"}},"typography":{"lineHeight":"1.2"}},"textColor":"tertiary","fontSize":"x-large","fontFamily":"cardo"} -->
```
MaggieCabrera: Let's use rems instead of px for the margins and paddings all over the pattern

patterns/testimonial.php
```diff
@@ -0,0 +1,27 @@
...
+  <p class="has-text-align-center has-tertiary-color has-text-color has-cardo-font-family has-x-large-font-size" style="margin-bottom:32px;line-height:1.2"><em><?php echo esc_html__('“Études has saved us thousands of hours of work and has unlocked insights we never thought possible.”', 'twentytwentyfour'); ?></em></p>
```
MaggieCabrera: The content of the paragraph should be separated from the borders of its container. It's hard to see in figma, but the horizontal spacing should be of 90px (5.625rem) on each side

MaggieCabrera (changes requested):
Thank you for your PR! just a few comments on it and it should be good to go

MaggieCabrera:
If you rebase the PR agains trunk you will get the latest changes for the color palette. This pattern should be using Contrast as a background color and base as a text color

patterns/testimonial.php
aplauche: Will do! Thanks

aplauche:
Hmm, added changes from review, but currently getting a generic "Can't use 'tar -xzf' extract archive file" on code checks? Was working earlier and caught a few formatting things that I fixed. Can look more into it a little later today.

MaggieCabrera (approved):

PR #47, opened 2023-08-24
Add code style config files

### LINKED ISSUE #5: Add code quality checks to PRs
We don't expect to have much CSS or PHP, but the code we do have should be checked against WP standards.

For reference, here's the same discussion for Twenty Twenty-Two: https://github.com/WordPress/twentytwentytwo/issues/24

vcanales:
Copy & paste from twentytwentythree to get started.

Ref. https://github.com/WordPress/twentytwentyfour/issues/5

### Testing instructions

1. Install node and composer, and then project dependencies:
```sh
npm install
composer install
```
2. Check that linting commands works. The ones available are:

- `npm run lint:css` lints and autofixes where possible the CSS
- `composer run analyze [filename.php]` statically analyzes PHP for bugs
- `composer run lint` checks PHP for syntax errors
- `composer run standards:check` checks PHP for standards errors according to [WordPress coding standards](https://developer.wordpress.org/coding-standards/)
- `composer run standards:fix` attemps to automatically fix errors

In this branch, there are no errors; to test, you could add errors, for example, in pattern files by messing with the indentation or spaces.


lint.yml
jeffpaul, line 1: This seems like a duplicate file to the one also being added in `.github/workflows`, so perhaps we can remove this file?

jeffpaul:
Nice work... passing checks, hooray!

mikachan (approved):
Nice, all the commands are working well for me!

PR #48, opened 2023-08-24
Centered Statement pattern

amedv:
Centered Statement pattern based on Figma template. Set to wide width and with paddings on all sides and text centered. Text translation split into three strings because of use <em> tag for name highlight. Original text in Figma written in Italian, used Google Translate to translate into English.


patterns/centered-statement.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Title: Centered Statement
+ * Slug: twentytwentyfour/centered-statement
+ * Categories: text
+ */
+
+?>
+
+<!-- wp:group {"align":"wide","layout":{"type":"default"}} -->
+<div class="wp-block-group alignwide"><!-- wp:paragraph {"align":"center","style":{"typography":{"lineHeight":"1.3","fontStyle":"normal","fontWeight":"400"},"spacing":{"padding":{"top":"12rem","bottom":"12rem","right":"13rem","left":"13rem"}}},"backgroundColor":"base","textColor":"contrast","fontSize":"large","fontFamily":"cardo"} -->
```
carolinan: Here, we shouldn't duplicate the values that are the defaults, like the font style, font weight, background color, and text color.
We also don't need to duplicate the settings and add them to both the group and the paragraph.

patterns/centered-statement.php
```diff
@@ -0,0 +1,22 @@
...
+    <?php
+        echo sprintf('%1$s<em>%2$s</em>%3$s',
+            esc_html__( 'These photographs not only captured the physical beauty of the ', 'twentytwentyfour' ),
+            esc_html__( 'Piva Arts &amp; Crafts Hotel', 'twentytwentyfour' ),
+            esc_html__( ', but also conveyed the emotions and feelings the hotel evokes in its guests', 'twentytwentyfour' ),
+        );
+    ?>
+</p>
+<!-- /wp:paragraph --></div>
+<!-- /wp:group -->
```
carolinan:
```suggestion
<!-- /wp:group -->

```
Add a blank line.

patterns/centered-statement.php
carolinan: Instead of hardcoding 12  and 13 rem,  please select a spacing preset.

MaggieCabrera (changes requested):
Besides the already pointed out issues, this PR needs some linting. Please check the [readme](https://github.com/WordPress/twentytwentyfour#testing) for instructions on how to do so. You will probably need to rebase this PR and run composer run standards:fix to do so. Let us know if you have any issues doing this.

MaggieCabrera:
I made some changes and switched the text back to Italian (the language was intended to be Italian, we shouldn't translate the patterns), somehow I see my changes in the fork but not on this PR, I'm going to give it some time

MaggieCabrera (approved):

PR #59, opened 2023-08-24
Update heading structure (Issue 54)

### LINKED ISSUE #54: The H1 is identical on every page, regardless of content due to Site Title block level
## Issue summary

The Site Title in the navigation bar or header template of the theme is wrapped in an `<h1>` element. This causes the H1 of every single page to be the same regardless of the content of the page. The first level heading on the page should be the **page title,** not the site title.
Screen Shot 2023-08-24 at 4 01 54 PM

## Current code
```HTML
<!-- wp:group {"style":{"spacing":{"blockGap":"0px"}}} -->
<div class="wp-block-group"><!-- wp:site-title /--></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```

## Steps to reproduce

1. Open up any page in your WordPress site with the TwentyTwentyFour theme
2. Inspect the site title in the left hand side of the header block
3. Observe the h1 wrapped around the site title.

## Recommendation

1. The site title block allows us to select the heading level used when outputted to the page. In this case, the level should be set to paragraph or level: 0. This will output the site title as a paragraph instead of a heading and it can then be styled according to the design spec. 
2. Amend each of the page templates so that the top level heading where the page title is outputted is an H1 and not an H2. 

### Code for files header.html, header-copy.html, header-writer.html, header-portfolio.html

```html
<!-- wp:group {"style":{"spacing":{"blockGap":"0px"}}} -->
<div class="wp-block-group"><!-- wp:site-title {"level":0} /--></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```

## Specifications
**Relevant Templates**
[header-copy.html](https://github.com/WordPress/twentytwentyfour/blob/trunk/parts/header-copy.html)
[header-portfolio.html](https://github.com/WordPress/twentytwentyfour/blob/trunk/parts/header-portfolio.html)
[header-writer.html](https://github.com/WordPress/twentytwentyfour/blob/trunk/parts/header-writer.html)
[header.html](https://github.com/WordPress/twentytwentyfour/blob/trunk/parts/header.html)

WCAG Principle: Perceivable
WCAG SC: 1.3.1  Info and Relationships

jessynd:
This pull request resolves #54

### DIFF
--- a/parts/footer-copy.html
+++ b/parts/footer-copy.html
@@ -7,7 +7,7 @@
 <p style="font-size:1.3rem">✦</p>
 <!-- /wp:paragraph -->
 
-<!-- wp:site-title /-->
+<!-- wp:site-title {"level":0} /-->
 
 <!-- wp:site-tagline /--></div>
 <!-- /wp:group --></div>
--- a/parts/footer-portfolio.html
+++ b/parts/footer-portfolio.html
@@ -11,7 +11,7 @@
 <p style="font-size:1.8rem">⊙</p>
 <!-- /wp:paragraph -->
 
-<!-- wp:site-title /-->
+<!-- wp:site-title {"level":0} /-->
 
 <!-- wp:site-tagline /--></div>
 <!-- /wp:group --></div>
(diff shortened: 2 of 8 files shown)


MaggieCabrera (approved):
Works as expected, thank you for the PR!

PR #64, opened 2023-08-25
Grid of Features Pattern

### LINKED ISSUE #108: Block Patterns - Grid of Features
**Description**

🛠️ The goal of this issue is to track the implementation of the "Grid of Features" pattern.
🎨 [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=536%3A1991&mode=design&t=LHSPHKUrPq556l59-1)

Grid of Features

maneshtimilsina:


kafleg (commented):
Would you please see the design and make the outline that is in the heading of this pattern?

maneshtimilsina:
@kafleg Please check it now.

kafleg (approved):
LGTM

kishanjasani:
@kafleg Code formatting isn't good to understand! Could we try to indent code so it's easy to read code.

MaggieCabrera (changes requested):
@kishanjasani indenting patterns is tricky business and can potentially break them, sometimes blocks that have inner blocks will break when it's indented. We can try it and see what happens but we need to keep that in mind.

@maneshtimilsina two things: this is a pattern that is contained in the home page, so we don't want to have the markup twice, the way to go about this is to remove the markup from the home and instead use the pattern block to reference it, so the code only lives in one place. The second thing is that after doing a quick scan of the marlup, I see there are differences between this pattern and what's on the home page: there are spacings that look different. Since this pattern has been already created by the designer, I would prefer if those values stay the way they were.
 
Thank you for the PR!

maneshtimilsina:
@MaggieCabrera changes are done and updated as per your instructions. Please verify it and let me know if there is anything left to do.

maneshtimilsina:
@MaggieCabrera I haven't replaced the content of `home.php` to avoid merge conflict. I think it will be better to replace the codes of `home.php` with patterns once all patterns are approved and merged to the trunk.

maneshtimilsina:
@MaggieCabrera Please review this PR and kindly let me know if there is anything left to do or if you want any changes.

MaggieCabrera (approved):
Thank you, I think this is good to go

PR #70, opened 2023-08-25
Created testimonial pattern

kafleg:
This pattern is created from home.php


carolinan:
Hi, this pattern has already been submitted here: https://github.com/WordPress/twentytwentyfour/pull/36.

PR #71, opened 2023-08-25
List custom templates

maneshtimilsina:
- Added a list of custom templates on the `theme.json`

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -1,799 +1,799 @@
 {
-	"settings": {
-		"appearanceTools": true,
-		"color": {
-			"palette": [
-				{
-					"color": "#f9f9f9",
-					"name": "Base",
-					"slug": "base"
-				},
-				{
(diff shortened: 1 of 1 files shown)


carolinan:
Hi
I think theme.json should also include the "single-writer" and "single-with-sidebar" templates?

maneshtimilsina:
@carolinan, I have added remaining templates (Single Writer and Single with Sidebar).

MaggieCabrera:
Is there a way to fix the PR so we can see just the lines that you change? the diff is kind of scary :D

maneshtimilsina:
> 

It is difficult to see diff in some cases. There might be a better way but for this file, I have added codes from lines 797 to 854. I think this will help.

maneshtimilsina:
Closing this PR in favor of https://github.com/WordPress/twentytwentyfour/pull/103/

PR #72, opened 2023-08-25
Replace variable font - Inter

maneshtimilsina:
- Replaced multiple files of variable font Inter with a single woff2 file.


beafialho:
The font `Inter` should be removed as it's there by mistake, the [correct font from the style guide](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=401-1691&mode=design&t=tQqcX6ylmUY0sScX-4) is a system sans serif font.

Issue #49, opened 2023-08-24
Sizing conformity for patterns — rems vs ems vs pixels


hiyascout:
Is a decision in order on block editor sizing units for both fonts and spacing, and/or setting variables?

I think expecting everyone to build first drafts in the same units might be a tall order, but we could aim to convert/conform patterns as we revise, and those of us who are comfortable working in variables/rems could do so from the start. 

In the mockup, here are the sizes: h1 48px, h2 38px, h3 28px, h4 24px, h5 20px, h6 18px, p 14px. (So would 1rem=14px?) 

Context: I'm building the Posts with Sidebar pattern and converting the pixels to rems where I think it works more consistently (horizontal text spacing for a post meta row, etc).

melchoyce:
Related: https://github.com/WordPress/twentytwentyfour/issues/34

hiyascout:
Thank you @melchoyce !   Also noting that in the pattern I'm working on specifically, there are some font sizes that are just short or just bigger than the presets.

richtabor:
Related as well, for spacing: https://github.com/WordPress/twentytwentyfour/pull/102

hiyascout:
If we follow @richtabor's Fibonacci proposal in #102, the fonts could be adjusted as written below. This might stretch past the original vision for the design (especially in that it has two sizes that are larger than the original 48 pt), and the use of two sequences might be convoluted. 

The original sequence (14, 18, 20, 24, 28, 38, 48) doesn't match any Fibonacci sequence, but we could alternate between two Fibonacci sequences (7 and 9). (I picked 7 and 9 because they seemed to have the closest numbers to the original sequence--v. arbitrary.) {But, 7+9 is 16 and there are 16 letters in twentytwentyfour 🤣}

Fibonacci sequence starting at 7 includes:
...14, 21, 35, 56...

Fibonacci sequence starting at 9 includes:
...18, 27, 45, 72...

Combined sequence could be:
14, 18, 21, 27, 35, 45, 56, 72

We could then use fluid font sizes based on this, using an average as the preferred size: 

h1 clamp([56], [64], [72]);
h2 clamp([45], [50.5], [56]);
h3 clamp([35], [40], [45]);
h4 clamp([27], [31], [35]);
h5 clamp([21], [24], [27]);
h6 clamp([18], [19.5], [21]);
p clamp([14], [16], [18]);

This would require that font sizes in patterns that people are building be converted to variables. Is that too far removed from the original vision?

[...]

Issue #34, opened 2023-08-24
Document design tokens


melchoyce:
In creating my first pattern from the Figma file, I encountered some decisions I wasn't sure how to approach:

- Font sizes in Figma that aren't in theme.json
- Default spacing units

We should document some of these tokens and how to use them when creating patterns.

luminuu:
Another one:
- Fonts

richtabor:
> Default spacing units

Proposal here for fluid spacing: https://github.com/WordPress/twentytwentyfour/pull/102. 

I'm thinking the `40` preset to be used as the standard padding around edges.

colorful-tones:
It is currently quite challenging to try and build a pattern. We should try to streamline some of these tokens and make them readily available. I recommend we update the variables in the Figma file _and_ make some docs (markdown table in README at the very least) for contributors to grab and reference key tokens. I'm happy to help and will be at tomorrow's meeting. 😄

richtabor:
> I recommend we update the variables in the Figma file. 

I've done that. What kind of table do you think we need? I think as long as you're using a spacing, color, font size preset, then you're sitting good.

colorful-tones:
I think [#213](https://github.com/WordPress/twentytwentyfour/pull/213) was a step in the right direction, but we'll have to be mindful of keeping it up to date if anything changes.

PR #86, opened 2023-08-26
Add "posts" pattern

### LINKED ISSUE #113: Block Patterns - Posts
**Description**

🛠️ The goal of this issue is to track the implementation of the "Posts" pattern.
🎨 [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=492%3A4728&mode=design&t=LHSPHKUrPq556l59-1)

Posts

melchoyce:
Tried replicating this pattern as closely as possible:

image
[Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=492-4728&mode=design&t=Sp21Ytn3WpNk8aWP-4)

Some issues I ran into:
- The spacing presets don't match the space in the mockup.
  - The preset I used for spacing between columns, block title, and post title is 36px, but the spacing in Figma is 32px
  - The preset I used for spacing between meta is off; there's not a great way to measure this since it's formatted as a single paragraph, but I'm using a row block to get all of the information in one line
- The font size presets didn't match the mockup (I used custom sizing for the heading since the presets were way off, and the "small" size for the excerpt because it was only 1px off)
- I couldn't easily add "by" before the author name without it being a separate block

These issues are related to #34 and #49

### DIFF
--- a/patterns/posts.php
+++ b/patterns/posts.php
@@ -8,7 +8,7 @@
 ?>
 <!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|80","bottom":"var:preset|spacing|80"}}},"layout":{"type":"constrained"}} -->
 <div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--80);padding-bottom:var(--wp--preset--spacing--80)"><!-- wp:heading {"align":"wide","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|60"}}}} -->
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
Hey @melchoyce thank you for this. I rebased the PR and updated the colors, but the spacings seem kind off to me after we changed the global values. Can you have another look at it?

MaggieCabrera:
oh, I will add translations to the strings too

luminuu:
I've updated the spacings on this one with the latest version of trunk.

MaggieCabrera (approved):

Issue #10, opened 2023-08-24
Style Variations


luminuu:
There will be three additional style variations available:

- [x] Default - Light Serif
- [ ] Dark - Serif
- [ ] Light - Sans-Serif
- [x] Dark - Sans-serif

melchoyce:
I'd love to design a variation for Twenty Twenty-Four! Is there a process for community members to submit ideas yet?

richtabor:
> I'd love to design a variation for Twenty Twenty-Four! Is there a process for community members to submit ideas yet?

I'm not one of the leads, but I'm sure PRs are welcome :)

richtabor:
We can probably come up with better names for these. 

Perhaps natural/organic color terms, like "Clay", "Onyx", "Ice", "Sesame", "Dove", "Jade", "Moss" etc. 

Reads a bit less tech and more human perhaps.

MaggieCabrera:
> > I'd love to design a variation for Twenty Twenty-Four! Is there a process for community members to submit ideas yet?
> 
> I'm not one of the leads, but I'm sure PRs are welcome :)

Always!

damen02:
Hello!
I have an idea that I think would be a great improvement on the use of style variations. I don't know if this is the right place to do it.

Currently it is possible to choose only one style variation for all website posts. It would be great to be able to apply a different style variation for different templates or CPTs. I mean changing typography, colors and spacing for each case depending on the type of content.

I can imagine many scenarios in which this would be very useful.
- Let's say you have a clothing store and you want to apply a style to products for babies, another for women and another for men.
- Or let's say you have a magazine and you want to apply a style to all the posts in the sports section, a different style to all the posts in the science section, and another for art, and so on.
- Or let's say you are simply an eclectic artist and you have different kind of works to offer, like paintings, sculpture or virtual.

The idea is not to limit the style variations to a single option, but to be able to use several on the same website, choosing according to the type of content.

Thank you for reading!

richtabor:
> Currently it is possible to choose only one style variation for all website posts. It would be great to be able to apply a different style variation for different templates or CPTs. I mean changing typography, colors and spacing for each case depending on the type of content.

Yes! I think that's an interesting idea for sure. 

Theme.json partials (where you can apply colors—and other styles—to specific areas of a site) is something being explored in https://github.com/WordPress/gutenberg/issues/40318 and https://github.com/WordPress/gutenberg/issues/48581.

PR #92, opened 2023-08-26
Style Variations

huzaifaalmesbah:
Basic style variations as #10

### DIFF
--- a/styles/dark-sans-serif.json
+++ b/styles/dark-sans-serif.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://schemas.wp.org/wp/6.2/theme.json",
+    "$schema": "https://schemas.wp.org/trunk/theme.json",
     "version": 2,
     "title": "Dark Sans-serif",
     "settings": {
--- a/styles/dark-sans-serif.json
+++ b/styles/dark-sans-serif.json
@@ -1,0 +1,42 @@
+{
+    "$schema": "https://schemas.wp.org/wp/6.2/theme.json",
+    "version": 2,
+    "title": "Dark Sans-serif",
+    "settings": {
+      "color": {
+        "palette": [
+            {
+                "color": "#222222",
+                "name": "Base",
+                "slug": "base"
+            },
+            {
+                "color": "#ffffff",
+                "name": "Contrast",
+                "slug": "contrast"
+            },
+            {
+                "color": "#222222",
+                "name": "Primary",
+                "slug": "primary"
+            },
+            {
+                "color": "#a2a2a2",
+                "name": "Secondary",
+                "slug": "secondary"
+            },
+            {
+                "color": "#303030",
+                "name": "Tertiary",
+                "slug": "tertiary"
+            }
+        ]
+      }
+    },
+    "styles": {
+        "typography": {
+          "fontFamily": "var:preset|font-family|inter"
+        }
+      }
+  }
+  
--- a/styles/dark-serif.json
+++ b/styles/dark-serif.json
@@ -1,0 +1,37 @@
+{
+    "$schema": "https://schemas.wp.org/wp/6.2/theme.json",
+    "version": 2,
+    "title": "Dark serif",
+    "settings": {
+      "color": {
+        "palette": [
+            {
+                "color": "#222222",
+                "name": "Base",
+                "slug": "base"
+            },
+            {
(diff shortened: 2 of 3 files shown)


styles/dark-sans-serif.json
```diff
@@ -0,0 +1,42 @@
...
+                "color": "#303030",
+                "name": "Tertiary",
+                "slug": "tertiary"
+            }
+        ]
+      }
+    },
+    "styles": {
+        "typography": {
+          "fontFamily": "var:preset|font-family|inter"
```
MaggieCabrera, line 38: Should we expand on this to include the different font weights @matiasbenedetto @mikachan ?

styles/dark-serif.json
```diff
@@ -0,0 +1,37 @@
+{
+    "$schema": "https://schemas.wp.org/wp/6.2/theme.json",
```
MaggieCabrera: This can probably be the 6.3 schema

styles/dark-serif.json
```diff
@@ -0,0 +1,37 @@
...
+                "name": "Base",
+                "slug": "base"
+            },
+            {
+                "color": "#ffffff",
+                "name": "Contrast",
+                "slug": "contrast"
+            },
+            {
+                "color": "#222222",
```
MaggieCabrera, line 19: I don't think we should be duplicating colors here, it's confusing for the user. Will we break patterns or templates if we remove it?

styles/dark-serif.json
huzaifaalmesbah: ok. i solved it

styles/dark-serif.json
huzaifaalmesbah: Will not break but will get contrast color from default style

styles/dark-serif.json
huzaifaalmesbah: I checked twentytwentythree using their trunk schema and also twentytwentyfour theme.json use trunk. Can I Updating schema from version 6.2 to trunk

styles/dark-sans-serif.json
matiasbenedetto: Nope, I don't think so because this part of the theme.json only sets the default preset for the elements. The Inter fonts weights were already created in `settings.typography.fontFamilies`.

styles/dark-serif.json
richtabor: Yea, we should not be duplicated color values. It's not helpful having two controls in the color palette UI for selecting the same color.

richtabor:
Noting that we'll need to adjust the palette based on the https://github.com/WordPress/twentytwentyfour/pull/106 proposal—when/if merged.

styles/dark-sans-serif.json
carolinan: Inter has been removed from the theme; please see https://github.com/WordPress/twentytwentyfour/pull/72#issuecomment-1697734544

huzaifaalmesbah: ok. and also need to color palette update

huzaifaalmesbah: Thanks for suggestion @carolinan

MaggieCabrera (changes requested):
I'm testing this and this is what I see:

- Dark sans-serif and dark serif look exactly the same. The font never changes because we are only defining it and not applying it to any blocks
- light sans-serif looks exactly the same as the default theme
- We need to add the extra colors that the default theme adds or we will be losing them from the palette

huzaifaalmesbah:
Thanks for the review and your feedback @MaggieCabrera  It's too old PR.
I will do it as soon as possible and update PR.

richtabor:
> Thanks for the review and your feedback @MaggieCabrera It's too old PR.

Agree. It may be best to close this and start a new one (with just one variation). But I'd suggest waiting a bit longer till we have more patterns/templates fleshed out (maybe by the end of this week). So you can properly test/adjust based on what's in the parent theme.json.

huzaifaalmesbah:
> Agree. It may be best to close this and start a new one (with just one variation). But I'd suggest waiting a bit longer till we have more patterns/templates fleshed out (maybe by the end of this week). So you can properly test/adjust based on what's in the parent theme.json.

Okay. I agree with you. I close this PR. I think we should work on it once it's all done.

Issue #74, opened 2023-08-25
Landmarks issues for Accessibility


00travelgirl00:
I have made a look at the landmarks and noticed following problems. See also handbook regarding landmarks: https://make.wordpress.org/themes/handbook/review/accessibility/required/#aria-landmark-roles 
 
homepage:
- main landmark is not present.
- not all content is inside a landmark 

pages:
- Titel / H1 is not in main

posts:
- Titel / H1 is not in main
- meta informations and comments are not part of any landmark. main landmark would be good

## suggestion
archive:
- wrap posts in `<article>` instead of divs

I'm not sure, if this last point is a requirement. But I have seen this practice with `<article>` several times and I think its a nice practice and user friendly for screen-readers.

graylaurenm:
I've worked on this for the homepage, pages, and archives. I haven't had a change to look at the posts yet. I will submit a pull request momentarily.

graylaurenm:
I noticed that `home-portfolio` and `home-writer` also need landmarks.

richtabor:
How’re we looking on this front? Did we over all the bases?

carolinan:
All templates are not complete yet and here are still problems.

Header: It was decided that the sticky header should be used everywhere. I think the wrapping group that makes the template part sticky should be the `<header>` element.
Current:
```
<div>
  <header>
  stuff
  </header>
</div>
<main></main>
<footer></footer>

```
Expected:

```
<header>
  <div>
  stuff
  </div>
</header>
<main></main>
<footer></footer>
```


404,, archive, home.html, index,  page-alternative, page-no-title, page-portfolio, page
Main landmark is present.
All content is inside a landmark

page-writer: For some reason the `<main> `is inside another group that uses `<div>.` These could be switched so that the `<main>` is the first group after the header. Let's make this consistent.

Search: has content outside landmarks. Fixed by #187

single-with-sidebar: Has content outside landmarks, missing landmark for the sidebar.

single-writer: Has content outside landmarks
single: Has content outside landmarks

home-writer: Has not been re-created using patterns yet and is missing landmarks.

MaggieCabrera:
yeah, the template work is still on going, in fact the header PR hasn't been merged yet, but this is a good recap to have, thank you

MaggieCabrera:
I think most of the templates are in now (only the portfolio ones are a WIP), let's have another pass and make sure we are covering all the landmarks

MaggieCabrera:
[...]

PR #95, opened 2023-08-26
Improve landmarks for accessibility

graylaurenm:
Begins work on recommendations made in #74. Specifically, looks at the homepage, pages, and archives but not the posts.

### DIFF
--- a/patterns/home.php
+++ b/patterns/home.php
@@ -6,7 +6,8 @@
  * Inserter: no
  */
 ?>
-<!-- wp:group {"style":{"spacing":{"padding":{"top":"6vh","bottom":"6vh"}}},"layout":{"type":"constrained"}} -->
+<!-- wp:group {"tagName":"main","layout":{"type":"default"}} -->
+<main class="wp-block-group"><!-- wp:group {"style":{"spacing":{"padding":{"top":"6vh","bottom":"6vh"}}},"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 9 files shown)


patterns/home.php
```diff
@@ -6,7 +6,8 @@
  * Inserter: no
  */
 ?>
-<!-- wp:group {"style":{"spacing":{"padding":{"top":"6vh","bottom":"6vh"}}},"layout":{"type":"constrained"}} -->
+<!-- wp:group {"tagName":"main","layout":{"type":"default"}} -->
+<main class="wp-block-group"><!-- wp:group {"style":{"spacing":{"padding":{"top":"6vh","bottom":"6vh"}}},"layout":{"type":"constrained"}} -->
```
MaggieCabrera, line 10: Can we edit the group instead that is existing instead of adding another one?

graylaurenm: That existing group is only for the first row of content. The new group puts all the non-header and non-footer content into the "main" landmark.

MaggieCabrera: that makes a lot of sense, I didn't read the code correctly, thank you

MaggieCabrera (approved):
Thank you for your work, I'll add you to the repo so you don't need to fork for future PRs

templates/archive.html
```diff
@@ -5,18 +5,20 @@
 	<!-- wp:query-title {"type":"archive","style":{"spacing":{"margin":{"bottom":"100px"}}}} /-->
 
 	<!-- wp:post-template -->
-		<!-- wp:group -->
-		<div class="wp-block-group">
+		<!-- wp:group {"tagName":"article"} -->
+		<article class="wp-block-group">
```
carolinan, line 9: Since this is repeated for each post; -for each individual result on the archive page, won't this create too many landmarks?

patterns/home.php
richtabor: Just noting that landmarks aren’t necessary in patterns, but they are in templates. Patterns are inserted many times over; we can assume users won’t have multiple patterns (i.e. multiple main tags).

PR #102, opened 2023-08-27
Add fluid spacing scale

richtabor:
**Description**

Add a consistent fluid spacing scale that we can start leveraging in patterns, instead of using arbitrary values. This proposal uses the Fibonacci sequence. 

We can continue to iterate and refine the values, but I think it'd be good to have a consistent set to start working from. 

**Testing Instructions**

1. Activate the theme. 
2. Add spacer block. 
3. Adjust the height to see the spacing presets. 
4. View on mobile to see the fluidity. 

**Screenshots**

Desktop: 
CleanShot 2023-08-27 at 10 56 52

Mobile: 
CleanShot 2023-08-27 at 11 04 14

Example of this fluid spacing expressed in a pattern, with `40` as the outer padding: 
CleanShot 2023-08-28 at 09 44 47
CleanShot 2023-08-28 at 09 44 58

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -115,13 +115,38 @@
 			"wideSize": "1280px"
 		},
 		"spacing": {
-			"spacingScale": {
-				"increment": 1.5,
-				"mediumStep": 1.5,
-				"operator": "*",
-				"steps": 7,
-				"unit": "rem"
-			},
+			"spacingSizes": [
+				{
+					"name": "1",
+					"size": "min(24px, 4vw)",
+					"slug": "10"
+				},
+				{
+					"name": "2",
+					"size": "min(36px, 6vw)",
+					"slug": "20"
+				},
+				{
+					"name": "3",
+					"size": "min(60px, 10vw)",
+					"slug": "30"
+				},
+				{
+					"name": "4",
+					"size": "min(96px, 16vw)",
+					"slug": "40"
+				},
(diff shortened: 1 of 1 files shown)


theme.json
```diff
@@ -115,13 +115,38 @@
...
+					"slug": "10"
+				},
+				{
+					"name": "2",
+					"size": "min(36px, 6vw)",
+					"slug": "20"
+				},
+				{
+					"name": "3",
+					"size": "min(60px, 10vw)",
```
MaggieCabrera: Should we use rems instead of px here?

richtabor: Good call, incoming.

richtabor: Done. Thanks!

MaggieCabrera (approved):
Love this, thank you!

beafialho:
These presets look good!

I'm seeing a lot of spacing that was not supposed to be there, particularly in stack blocks in the home template. This may have happened everywhere else. It seems that it may have inheriting one of the presets, but one that's way too high? Here's what I mean, comparing it to a demo site I created with the intended spacings:

https://github.com/WordPress/twentytwentyfour/assets/65220155/097cc565-5fec-4b68-8079-3f893cc863e1

richtabor:
Yea, we need to adjust the templates to use the system.

PR #105, opened 2023-08-28
Remove default color palette

richtabor:
### Description

Proposing that we remove the default color palette, in support of making the Twenty Twenty Four experience much more tailored. It's less distracting without them, and we don't use those values anyhow. 

We'll continue to refine the theme colors, but in the meantime the defaults seem out of place. 

### Screenshots

| Before  | After |
| ------------- | ------------- |
|CleanShot 2023-08-28 at 09 50 13|CleanShot 2023-08-28 at 09 49 44|

### Testing

1. Activate the theme.
2. Add a page with a paragraph.
3. Select "Color" to see the default colors not displaying.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -2,6 +2,7 @@
 	"settings": {
 		"appearanceTools": true,
 		"color": {
+			"defaultPalette": false,
 			"palette": [
 				{
 					"color": "#f9f9f9",


melchoyce:
Agreed 👍

jeffikus:
+1 from me, any thoughts @MaggieCabrera @luminuu ?

jeffikus (approved):
I'm fine with this, but I think it would be good to get feedback from the theme leads before we merge this. Approving it for now though.

MaggieCabrera (approved):
I like this idea

PR #109, opened 2023-08-28
Add duotone presets

richtabor:
**Description**
Adds duotone presets based on the existing beautiful color palette.  Also removes the default duotones, to promote a more cohesive design experience with the theme. 

Aside: Is is possible to reference the color presets, instead of the hex values, in duotone JSON?

**Screenshots**

CleanShot 2023-08-28 at 13 13 00

CleanShot 2023-08-28 at 13 15 39

**Testing Instructions**

1. Activate the theme.
2. Add an image.
3. See duotone presets.


MaggieCabrera:
> Aside: Is is possible to reference the color presets, instead of the hex values, in duotone JSON?
> 

Sadly no :(

theme.json
```diff
@@ -2,6 +2,49 @@
...
 		"color": {
+			"defaultDuotone": false,
+			"duotone": [
+				{
+					"colors": [
+						"#111111",
+						"#ffffff"
+					],
+					"slug": "black-and-white",
+					"name": "Duotone 1"
```
MaggieCabrera: I think this would make more sense to follow the slug, since it's the one the user will see

richtabor: Good catch! Done.

theme.json
```diff
@@ -2,7 +2,50 @@
...
+					"slug": "black-and-white",
+					"name": "Black and white"
+				},
+				{
+					"colors": [
+						"#111111",
+						"#C2A990"
+					],
+					"slug": "duotone-1",
+					"name": "Duotone 1"
```
MaggieCabrera: shall we do it for all of them? 🙈

richtabor: No, I don’t think “Accent 1 and Contrast” is very helpful. You won’t remember that; just the applied tones.

richtabor: And I think it’s better to keep slugs generic, so they’re more interoperable.

MaggieCabrera: I think that would be the way to go if duotone presets were actually dynamic but since they are not, a more descriptive name would be more useful for the user (black and red, rather than accent 1 and contrast)

joedolson: For a user with visual impairments, a descriptive color term is significantly more useful than an abstraction.

richtabor: @joedolson Is “Accent 1 / Base” helpful?

richtabor: @MaggieCabrera If duotones values are ever editable/dynamic, we’ll be in a scenario where it will read “Red and Black” but could be any other color values. 🫤

I think we can expect to be so eventually for sure.

joedolson: @richtabor Not really.  It doesn't give any context unless you have visual recent, in my opinion.

richtabor: > @richtabor Not really. It doesn't give any context unless you have visual recent, in my opinion.

Yea, I agree. It's not more helpful than "Duotone 1". 

The predicament we're in, is that we don't have duotone that can be edited outside of code, so colors are hard-coded (for now). When we do have the affordance to create and edit duotones—like you can colors—everyone with the theme installed will have color-specific labels for duotones, where the colors could be really any value. 

What do you think?

carolinan: Nobody is actively working on that feature that we know of.

If we call it "red" and change that color in the global style variations, then we only need to include the duotone in the variation file too.

richtabor: Made the change — thanks for chiming in!

joedolson (approved):

PR #127, opened 2023-08-28
Update README.md with design guidance and extras

richtabor:
- Removes some duplicated bits.
- Adds image header.
- Adds design guidance for contributors.
- Adds links to issue categories.


MaggieCabrera (approved):
Thank you so much, Rich

PR #136, opened 2023-08-29
Add offset grid posts pattern

### LINKED ISSUE #120: Block Patterns - Offset Grid of Image Posts
**Description**

🛠️ The goal of this issue is to track the implementation of the "Offset Grid of Image Posts" pattern.
🎨 [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=565%3A10502&mode=design&t=LHSPHKUrPq556l59-1)

Offset Grid of Image Posts

jordesign:
**Description**

First draft to add Offset Grid Posts pattern, and replace 'Home Portfolio' content with pattern
Addresses #120 

**Screenshots**

Desktop:
Screenshot 2023-08-29 at 10 18 06 pm

Mobile:
Screenshot 2023-08-29 at 10 18 33 pm

----
Note: we need to decide how to deal with the irregular spacer blocks on smaller screens. I've still issued this PR as a first draft but keen to know what folks think about once we go to a single column. We _could_ use CSS to remove those spacers, but I know we're trying to avoid that.


MaggieCabrera:
@richtabor you mentioned you wanted to have a look at this one, I would love your opinion on how to handle the spacing here. I wonder if we should have two versions of this: one that is a query loop and is inserted in the template that uses this, and one that is just a set of images.

jordesign:
Could we have a style variation for a Gallery block that alternated top/bottom margins for the items?

MaggieCabrera:
> Could we have a style variation for a Gallery block that alternated top/bottom margins for the items?

I'm not sure if it will work like the design, but you are welcome to try :D

jordesign:
Just noting/realising this would need to be a style variation for the Query Loop... Something I'm looking into..

My findings so far...

* Having a variation for the Query loop would allow us to programatically adding top padding to columns - but it would be in a uniform pattern (every second column of the row). And we'd also need 2 different variations for the different column blocks which alternate different rows.

* I tried using vertical alignment instead of padding - but the result was too _little_ gap and variation between items.

To my mind - what we essentially need to do is find a way to have the spacer blocks hidden on smaller screens, once we enter a single column. But I'm not sure what the most correct way to do that would be.

MaggieCabrera:
>     * Having a variation for the Query loop would allow us to programatically adding top padding to columns - but it would be in a uniform pattern (every second column of the row). And we'd also need 2 different variations for the different column blocks which alternate different rows.

I'm not sure this can ever be a variation for the query loop since it's using multiple instances of the block in different ways. 
 
> To my mind - what we essentially need to do is find a way to have the spacer blocks hidden on smaller screens, once we enter a single column. But I'm not sure what the most correct way to do that would be.

I have some ideas that I'm going to test shortly, I'll let you know if something sticks

MaggieCabrera:
@richtabor @colorful-tones I would love your opinion on this. Using min/max/clamp didn't work for 0 values here (or I couldn't make it work), do you know a better way to make the spacers 0 height on smaller screens? since they are not consistently between each item, the height needs to be 0 when we go to one column or the spacing looks uneven.

richtabor:
Yea, I'll take another look. We may need to hardcode min/max/clamp values for spacing in the spacers themselves — instead of using presets.

scruffian:
I tried a different approach here: https://github.com/WordPress/twentytwentyfour/pull/274

jordesign:
Closing this PR as the issue is resolved with #274

PR #137, opened 2023-08-29
Add presets to landing page CTA pattern

richtabor:
**Description**
Replaces arbitrary values with design presets. 

**Testing Instructions**

1. Search for "landing".
2. Add the pattern to the page.


carolinan (approved):

carolinan:
A heading for this might not be ideal, depending on where the user places the pattern on their page. But let's fix the presets for now.

PR #138, opened 2023-08-29
Fix accent-3 to match Figma

richtabor:
**Description**
Fix color for accent-3, to map to Figma. 

**Screenshots**

Before: 
CleanShot 2023-08-29 at 12 11 48

After: 
CleanShot 2023-08-29 at 12 11 28


huzaifaalmesbah (approved):
That's good

PR #145, opened 2023-08-30
Fixed name of theme

kafleg:
Instead of Twenty Twenty Four, the theme name should be Twenty Twenty-Four and changed in 3 different files.


huzaifaalmesbah (approved):
Thanks

alaminfirdows:
Great!

Issue #12, opened 2023-08-24
Add links to footer menu/items?


jordesign:
In the current Footer Template Part the columns in the right _appear_ as though they are a menu - but are just paragraph blocks. Should these be changed to be linked in some way? Or are they more decorative?

<img width="1072" alt="Screenshot 2023-08-24 at 8 28 27 pm" src="https://github.com/WordPress/twentytwentyfour/assets/1842363/624cc8a0-78f1-4275-993d-a0e31ef40140">

michaelshowes:
I will tackle this task

lefty55104:
@michaelshowes For accessibility, these links should be in a list rather than individual paragraphs. 

WCAG Principle: Perceivable
WCAG SC: 1.3.1 Info and Relationships

mrwweb:
I would think the links should be in a Navigation Block for the accessibility semantics @lefty55104 mentions + the general Block Editor semantics of "use a Navigation block to make a list of links that are site navigation".

Things that brings up:

1. The two-column style would require a custom block style probably, but that seems like it would be a great thing for a default theme to demonstrate!
2. What is the placeholder state of a navigation block when inserted? How can that be made as easy as possible for users?

carolinan:
And what is its mobile design? Not all users are going to limit it to two columns of 3.

jordesign:
Does having 2 columns with a nav block in each allow for mobile better - ending up stacking on top of each other?

carolinan:
What if it was one latest posts block and one navigation block?

carolinan:
Where in Figma is the two-column menu footer design? I can't find it.

jordesign:
Great question - I don't see them in the Figma at all...

So maybe the template part just needs to be corrected.

mrwweb:
> Does having 2 columns with a nav block in each allow for mobile better - ending up stacking on top of each other?

That would still probably be an accessibility issue to have two lists of three links rather than one list of six links.

joedolson:
This is something it would be nice to have a solution for; it's a problem I've also flagged in the wordpress.org footer.

carolinan:
@beafialho Is the navigation in the default footer horizontal like in Figma, or is it two columns like above?

beafialho:
[...]

PR #147, opened 2023-08-30
Fix designed with text of the footer

maneshtimilsina:
**Description**

Fixed the `Designed with` text's visibility issue.


carolinan:
I recommend against this solution because the text needs to be inside block patterns to be translation-ready.
There is a footer.php pattern which should be reused where possible. Where it can't be reused; new patterns should be created.

Depending on the result of these discussions the footers may change further:
https://github.com/WordPress/twentytwentyfour/issues/12
https://github.com/WordPress/twentytwentyfour/pull/14
https://github.com/WordPress/twentytwentyfour/pull/51

maneshtimilsina:
I did it for a quick fix but realized that the text on the footer needs to be inside the footer block to be translation-ready after you mentioned it. So I am closing this PR.

PR #148, opened 2023-08-30
Fixes #146 Footer text "Designed with" is not visible.

### LINKED ISSUE #146: Footer text "Designed with" is not visible.
**Description**
While testing Twenty Twenty Four theme, I found that the Footer text "Designed with" is not visible.

footer-text-not-visible

mohsinrasool:
Fixes #146 Footer text "Designed with" is not visible.

Fixed the issue by removing the _has-base-2-color_ css class as suggested.


huzaifaalmesbah:
You can check #147 
already discussed this issue.

PR #149, opened 2023-08-30
Update columns pattern with presets

richtabor:
**Description**
Uses presets instead of arbitrary values. 

**Screenshots**
CleanShot 2023-08-30 at 13 46 25


MaggieCabrera (approved):
LGTM, thank you Rich

PR #150, opened 2023-08-30
A11y fix: adds a checkmark block style for lists and replaces it in `home.php`

### LINKED ISSUE #46: Home page pattern - paragraphs are used for lists
## Issue summary

In the home template there is a design element that looks like a list but is contained in a div with paragraph tags. This is problematic because: 
1. This doesn't convey to Screen Readers that the content is in a list or the number of list items in each list. 
2. Checkmarks are read out loud as "checkmark" by the screenreader. 

Screen Shot 2023-08-24 at 2 26 09 PM

## Current Code
``` HTML
<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|30"}},"layout":{"type":"flex","orientation":"vertical"}} -->
<div class="wp-block-group"><!-- wp:paragraph -->
<p><strong>✓</strong> Dive into a world of thought-provoking articles</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p><strong>✓</strong> Read case studies that celebrate the artistry of architecture</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p><strong>✓</strong> Gain exclusive access to design insights</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group --></div>
```
## Steps to reproduce

1. Go to the home page or template of the theme. 
2. Scroll down to section with lists. 
3. Inspect the code for each list item to see the paragraph tags used

## Recommendation
1. Define a custom style of the default List Block
2. Use CSS to change the `list-style-type` property for unordered lists to use checkmarks. 
4. Swap out the default markup on the home pattern so it uses this style variation of the list block instead of the paragraphs. 

## Specifications
Template File:  Lists for lists, ol , ul , and dl used appropriately
**WCAG Principle:** Lists WCAG SC: 1.3.1 Info and Relationships
**Related File in the theme:**[ home.php](https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/home.php)

aurooba:
Closes #46.

**Description**

Right now there's a list pattern used in the home template for TT4 that uses paragraphs with checkmarks to simulate a list. This is an accessibility issue as described in #46:

> 1. This doesn't convey to Screen Readers that the content is in a list or the number of list items in each list.
> 2. Checkmarks are read out loud as "checkmark" by the screenreader.

This PR registers a new custom block style for Lists called `Checkmarks` to fix that issue.

**Screenshots**

Desktop view:

desktop view of the new checkmark list style in place on the homepage template

Mobile view:
mobile view of the new checkmark list style in place on the homepage template

Design discussion/question:
- the original design has considerable gap between the list items since they were paragraphs. Do we want to add spacing between list items when they are in this checkmark style or leave it be?

**Testing Instructions**

1. Activate the theme. -->
2. Create a new post or page.
3. Add a list and choose the checkmark list style.

### DIFF
--- a/patterns/home.php
+++ b/patterns/home.php
@@ -180,7 +180,7 @@
 <!-- /wp:heading --></div>
 <!-- /wp:group -->
 
-<!-- wp:list {"style":{"spacing":{"padding":{"right":"0","left":"0"}}},"className":"is-style-checkmark-list"} -->
+<!-- wp:list {"style":{"spacing":{"padding":{"right":"0","left":"var:preset|spacing|10"}}},"className":"is-style-checkmark-list"} -->
 <ul class="is-style-checkmark-list" style="padding-right:0;padding-left:0"><!-- wp:list-item -->
 <li>Collaborate with fellow architects</li>
 <!-- /wp:list-item -->
@@ -224,7 +224,7 @@
 <!-- /wp:heading --></div>
 <!-- /wp:group -->
 
-<!-- wp:list {"style":{"spacing":{"padding":{"right":"0","left":"0"}}},"className":"is-style-checkmark-list"} -->
+<!-- wp:list {"style":{"spacing":{"padding":{"right":"0","left":"var:preset|spacing|10"}}},"className":"is-style-checkmark-list"} -->
 <ul class="is-style-checkmark-list" style="padding-right:0;padding-left:0"><!-- wp:list-item -->
 <li>Dive into a world of thought-provoking articles</li>
 <!-- /wp:list-item -->
--- a/patterns/home.php
+++ b/patterns/home.php
@@ -180,8 +180,8 @@
 <!-- /wp:heading --></div>
 <!-- /wp:group -->
 
-<!-- wp:list {"style":{"spacing":{"padding":{"right":"0","left":"var:preset|spacing|10"}}},"className":"is-style-checkmark-list"} -->
-<ul class="is-style-checkmark-list" style="padding-right:0;padding-left:0"><!-- wp:list-item -->
+<!-- wp:list {"style":{"spacing":{"padding":{"right":"0","left":"var:preset|spacing|10"},"margin":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30"}}},"className":"is-style-checkmark-list"} -->
+<ul class="is-style-checkmark-list" style="margin-top:var(--wp--preset--spacing--30);margin-bottom:var(--wp--preset--spacing--30);padding-right:0;padding-left:var(--wp--preset--spacing--10)"><!-- wp:list-item -->
 <li>Collaborate with fellow architects</li>
 <!-- /wp:list-item -->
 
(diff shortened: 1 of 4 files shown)


carolinan:
Do we not need a separate block pattern for this?

aurooba:
@carolinan It's a change in two existing block patterns `Left Aligned Call to Action with Image` and `home`. I didn't see an existing block pattern just for the checkmark list style.

If we wanted to offer an easy to use checkmark list, a block variation might be a better fit perhaps?

carolinan:
I don't see any updates made to left-aligned-cta-image.php?

aurooba:
@carolinan Ooh, that's right, my bad. The original `left-aligned-cta-image.php` didn't have any checkmarks in it (although I thought it would be a good fit) so I never committed that.

Here's what `left-aligned-cta-image.php` looks like currently with a regular unordered list:

Screen Shot 2023-08-30 at 5 42 14 PM

Here's what it could look like with the checkmarks style:
Screen Shot 2023-08-30 at 5 42 23 PM

Should we get design feedback on this?

carolinan:
I found the pattern in Figma, and it has checkmarks:
https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=565-10465&mode=design&t=2hmKthGTssvnJpqM-0

aurooba:
Thanks for linking to the pattern in the design file directly, @carolinan. Much appreciated. The pattern has been updated, and I also overall tweaked the styling of checkmark lists globally so they better match the design:

Screen Shot 2023-08-31 at 9 41 28 AM

patterns/left-aligned-cta-image.php
```diff
@@ -13,17 +13,17 @@
 <h2 class="wp-block-heading"><?php echo esc_html__( 'Enhance your architectural journey with the Études Architect App', 'twentytwentyfour' ); ?></h2>
 <!-- /wp:heading -->
 
-<!-- wp:list {"style":{"typography":{"lineHeight":"1.75"}}} -->
-<ul style="line-height:1.75"><!-- wp:list-item -->
-<li><?php echo esc_html__( 'Collaborate with fellow architects', 'twentytwentyfour' ); ?></li>
+<!-- wp:list {"style":{"spacing":{"padding":{"left":"var:preset|spacing|10"},"margin":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30"}}},"className":"is-style-checkmark-list"} -->
+<ul class="is-style-checkmark-list" style="margin-top:var(--wp--preset--spacing--30);margin-bottom:var(--wp--preset--spacing--30);padding-left:var(--wp--preset--spacing--10)"><!-- wp:list-item -->
+<li>Collaborate with fellow architects</li>
```
carolinan, line 18: The translation functions need to be kept and preferably updated to use context. Example:
`<?php echo esc_html_x( 'Collaborate with fellow architects', 'Sample list content', 'twentytwentyfour' ); ?>`

patterns/home.php
```diff
@@ -180,19 +180,19 @@
...
-<!-- /wp:group --></div>
+<!-- wp:list {"style":{"spacing":{"padding":{"right":"0","left":"var:preset|spacing|10"},"margin":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30"}}},"className":"is-style-checkmark-list"} -->
+<ul class="is-style-checkmark-list" style="margin-top:var(--wp--preset--spacing--30);margin-bottom:var(--wp--preset--spacing--30);padding-right:0;padding-left:var(--wp--preset--spacing--10)"><!-- wp:list-item -->
+<li>Collaborate with fellow architects</li>
+<!-- /wp:list-item -->
+
+<!-- wp:list-item -->
+<li>Showcase your projects</li>
+<!-- /wp:list-item -->
+
```
carolinan, line 191: This section should be replaced with the matching pattern. We can do that for all patterns afterward as a separate PR, or it can be fixed here.

aurooba:
Noting that we need to incorporate an RTL solution as well as per the slack conversation today with @MaggieCabrera

MaggieCabrera:
the pattern Features with Images also needs the same implementation

MaggieCabrera:
https://github.com/WordPress/twentytwentyfour/pull/292 is working on adding some design adjustments after the pattern was reviewed. The markup provided in the adjustment turns the list into a paragraph block, which is not what we want, but it's the look we are aiming for. Let's keep the spacing, line height and layout from those adjustments when we make changes to the markup here.

MaggieCabrera:
I tried rebasing, had a hard time, figured it was better to just start over. Closing in favor of https://github.com/WordPress/twentytwentyfour/pull/369, let's continue there, but I basically picked up the changes from here and moved them to trunk manually with a couple of tweaks

PR #168, opened 2023-08-31
Update the default footer

carolinan:
**Description**
Updates the default footer using a site logo instead of the paragraph and a horizontal navigation block instead of the fake menu with two columns.
Uses the footer.php block pattern for the footer credit text.

**Testing Instructions**
Reset any saved changes to templates and template parts.
If needed, assign a menu to the navigation block in the footer.
If needed, assign a site logo to the site logo block.
View the updated footer in the Site Editor and on the front.


alaminfirdows:
Looks good, but I think font size need to be changed. On Figma it's guided to `12px` but currently it `17px` on the pattern.

carolinan:
12px is not accessible, so it can not be used. I used the smallest font size in theme.json.

alaminfirdows:
> 12px is not accessible, so it can not be used. I used the smallest font size in theme.json.

Can't we use the tag line's font size?

alaminfirdows (approved):
LGTM

huzaifaalmesbah (approved):

PR #172, opened 2023-08-31
Hide the asterisk from screen readers as a decorative element

### LINKED ISSUE #41: Decorative element being read by screen readers as asterisk
## Issue summary

Decorative element is reading as "asterisk" on screen readers

## Steps to reproduce
Implement voice over on macbook pro in chrome.

The decorative element in the below screen shot of the homepage is reading as "asterisk"

Screenshot 2023-08-24 at 3 06 20 PM

## Recommendation
Asterisk should be an image with no alt text vs just text.

## Specifications
Template File: Homepage
**WCAG SC:** [1.1.1](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)

aurooba:
**Description**

Closes #41. Adds `aria-hidden="true"` to all instances of the `✳︎` in the `home.php` template, so it doesn't get announced by a screen reader or assistive device.

**Testing Instructions**

1. Turn on VoiceOver on MacOS.
2. Read the `home.php` page on the front end
3. Notice that the "asterisks" don't get announced.


carolinan:
Hi, this has already been attempted. It does not work because the unexpected attribute causes a problem with the markup when saving.
See https://github.com/WordPress/twentytwentyfour/pull/51

MaggieCabrera:
Let's keep the conversation going on #51, thanks!

PR #176, opened 2023-09-01
Convert font files to woff2

LittleBigThing:
Switch from `ttf` files to `woff2` for the fonts. 

Woff2 is a format specific for web fonts. It is much lighter than ttf and woff. For example, the regular font file was reduced from 391 to 146 kb or 63% (unzipped).
Woff2 has great support in modern browsers and this format has been used in default themes for some time now.


LittleBigThing:
First take to fix #177

carolinan (approved):

Issue #27, opened 2023-08-24
Design Accessibility Improvements


anphira:
For people with Irlen syndrome, too high a contrast is an issue. Using #000 with #fff is not recommended as large groups of users (especially older) complain about the harshness on their eyes. Recommend softening the blacks to #222. 

On the links in post meta information - the post author and post category, recommend adding underline to show that they are links. While technically you can meet success criteria with a 3:1 ratio on the link text to surrounding text, it is strongly recommended to use clear underlines to make identification of links easier and more consistent. Additionally in the footer, in the "Designed with WordPress" where "WordPress" is the link, underline the link. 

While WCAG does not specify a minimum font size, the generally agreed minimum recommended size is 14px. Please replace all 12px text with 14px. Especially with relation to the large h2 font size, the 12px feels particularly small. 

Part of #3

MaggieCabrera:
@iamtakashi can you have a look at this feedback, see what kind of design changes we should make to address it?

joedolson:
I'll also note that underlined links in content regions is a requirement for the `accessibility-ready` tag in the theme directory. Regions like post meta data are a bit ambiguous on that point, but in my opinion they aren't obvious groups of navigation elements, which makes them more like content.

MaggieCabrera:
> I'll also note that underlined links in content regions is a requirement for the `accessibility-ready` tag in the theme directory. Regions like post meta data are a bit ambiguous on that point, but in my opinion they aren't obvious groups of navigation elements, which makes them more like content.

When are links "allowed" not to be underlined? Because commonly the header links are rarely ever underlined unless it's a specific design decision, for example. I think it would be better to have a blanket rule of "underline links everywhere except here and here", do you agree? also @melchoyce for your opinion here

joedolson:
Any link that's in a clearly demarcated region of navigation where there other affordances that indicate that the links are interactive is a candidate for non-underlined. It's not a firm guideline of what qualifies and doesn't qualify, because it's highly context and design sensitive. 

Links must be underlined when they're embedded within other text, such as within the body content of an article or comment. Areas like main navigation almost never require a link. Areas such as post meta data can fall into a broad spectrum of importance, depending on other clarifying indicators. I think we, as WordPress developers, tend to automatically identify post meta items such as author and category links as links, because we know internally how these are structured and what kind of content is there. Whether the design supports that is much more subjective. 

I'm not sure what the reference context for this issue is, exactly, so it's hard to make more specific comments.

[...]

Issue #111, opened 2023-08-28
Block Patterns - Pricing


juanfra:
**Description**

🛠️ The goal of this issue is to track the implementation of the "Pricing" pattern.
🎨 [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=492%3A4275&mode=design&t=LHSPHKUrPq556l59-1)

![Pricing](https://github.com/WordPress/twentytwentyfour/assets/252415/0cbf8057-749b-4d34-81dc-b58d5775c022)

PR #181, opened 2023-09-02
Update "Contrast" to #222222

carolinan:
**Description**

Update "Contrast" to #222222.
See https://github.com/WordPress/twentytwentyfour/issues/27#issuecomment-1699363393

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -63,7 +63,7 @@
 					"slug": "base-3"
 				},
 				{
-					"color": "#000000",
+					"color": "#222222",
 					"name": "Contrast",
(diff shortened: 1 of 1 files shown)


huzaifaalmesbah (approved):
Thanks For Update

richtabor:
This longer passes a11y (and other color variants barely do) — we need to be a bit more cognizant of these broad strokes.

I'd say #111 is a happy medium ground, where the contrast on colored elements is appropriate, but it's also not so start on the white background. 

CleanShot 2023-09-05 at 15 48 29

PR #187, opened 2023-09-04
Add H1 to templates

carolinan:
**Description**

Adds new H1 headings to templates that are missing them.
Update the heading level order where needed.

Partial for https://github.com/WordPress/twentytwentyfour/issues/54

**Testing Instructions**
View the update templates and check that the heading level is correct, that the H1 is present, and no heading levels are skipped.

**Note:**
If a template has a comments area, but the post or page does not have any comments, the heading level will skip from H1 directly to H3. The only way to work around this is to add a new H2 heading above the comments area.
This is because there is no way (that I know of) to change the level of the "Leave a reply" heading in the **Comments form** block. This extra heading is not part of the design, so I have left it as is for now, until it has been discussed...


templates/index.html
```diff
@@ -2,6 +2,9 @@
 
 <!-- wp:query {"tagName":"main","layout":{"type":"constrained"}} -->
 <main class="wp-block-query">
+	<!-- wp:heading {"level"1} -->
+	<h1 class="wp-block-heading">Watch, Read, Listen</h1>
```
carolinan: We need an H1 on the index, so I added the text that is found above the blog in the home template.
But this text is not translation-ready. Do we really want to add a new pattern only for a heading..?

MaggieCabrera: there's already [a pattern](https://github.com/WordPress/twentytwentyfour/pull/86) for this

carolinan: Will update this PR once that is merged

MaggieCabrera: done!

MaggieCabrera:
We have a comments pattern that we are not using (I have just submitted a PR to do just that). What does TT3 and TT2 do about the headings for those pages?

carolinan:
> We have a comments pattern that we are not using (I have just submitted a PR to do just that). What does TT3 and TT2 do about the headings for those pages?

In TT3 there is an extra heading above the comments area
In TT2 there is no extra heading above the comments area, so a heading level is skipped. (There are also two H1's on the single template)

carolinan:
I am still traveling, so feel free to continue with this or close it and start over without merge conflicts.

templates/single.html
```diff
@@ -48,4 +48,4 @@
 <!-- /wp:group --></div>
 <!-- /wp:group -->
 
-<!-- wp:template-part {"slug":"footer","tagName":"footer"} /-->
```
carolinan: Not sure what happened here, this change is unrelated, I copied the whole file from trunk to solve the merge conflict.

carolinan:
I ran into PHP errors testing some of the other templates on WordPress 6.4 alpha, so this is a partial fix only.

templates/index.html
```diff
@@ -1,32 +1,7 @@
...
-		<!-- wp:query-pagination -->
-			<!-- wp:query-pagination-previous /-->
-			<!-- wp:query-pagination-numbers /-->
-			<!-- wp:query-pagination-next /-->
-		<!-- /wp:query-pagination -->
-	</div>
-	<!-- /wp:group -->
+<!-- wp:group {"tagName":"main","align":"full","layout":{"type":"constrained"}} -->
+<main class="wp-block-group alignfull">
+	<!-- wp:pattern {"slug":"twentytwentyfour/posts"} /-->
```
carolinan, line 4: Because this pattern needs to have an H1 on the index, it means that other pages that use this pattern can end up with more than one H1.

MaggieCabrera: does that text need to be a heading at all?

carolinan: The index has to have an H1. I was told to use the pattern, so I needed to update that from H2 to H1.

carolinan: I think it is better to have two H1 than none.

MaggieCabrera: yeah, the text needs to be in a pattern anyway for translation. Let's go with this then

luminuu (approved):
LGTM!

PR #192, opened 2023-09-04
Block patterns: naming, categories and markup review

MaggieCabrera:
**Description**

This PR is a first pass at the already merged patterns to pick up a few things that have escaped reviews such as: IDs on image and query blocks, incorrect naming/labeling or categories, some markup changes such as using preset spacings whenever possible.

I did not fix the markup for the home template since that is still being worked on.

**Testing Instructions**

Check that all the patterns are inserted correctly after this PR with no issues. The sidebar writer pattern should appear on the Home writer template, and the Sidebar pattern should appear on the Single with sidebar template.

When you create a blank new page, the home pattern should popup as an option to populate the content with

### DIFF
--- a/parts/footer.html
+++ b/parts/footer.html
@@ -19,4 +19,8 @@
 </div>
 <!-- /wp:group -->
 
-<!-- wp:pattern {"slug":"twentytwentyfour/footer"} /-->
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group">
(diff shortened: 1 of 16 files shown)


richtabor:
I think this is going to conflict a lot with existing PRs with such wide changes. 😅

MaggieCabrera:
> I think this is going to conflict a lot with existing PRs with such wide changes. 😅

I was expecting to get it merged faster 🙈

PR #195, opened 2023-09-04
Added Translator Context for Long Paragraph

### LINKED ISSUE #178: Need to Use Translator Context for longer paragraph
Need to Use Translator Context for longer paragraph 

Use esc_html_x Instead of esc_html__ in long paragraph

**Files :** 
patterns/columns.php
patterns/home.php
patterns/sidebar-writer.php

shail-mehta:
Fixes https://github.com/WordPress/twentytwentyfour/issues/178


shail-mehta:
Created New PR for https://github.com/WordPress/twentytwentyfour/issues/178

patterns/sidebar-writer.php
```diff
@@ -64,17 +63,17 @@
...
 
 <!-- wp:group {"style":{"spacing":{"blockGap":"0.6rem"}},"layout":{"type":"flex","orientation":"vertical"}} -->
 <div class="wp-block-group"><!-- wp:paragraph {"style":{"typography":{"fontSize":"0.9rem"}}} -->
-<p style="font-size:0.9rem"><a href="https://tt4writer.mystagingwebsite.com"><?php echo esc_html__( 'Dernier rapport sur l\'inflation', 'twentytwentyfour' ); ?></a><?php echo esc_html__( '↗', 'twentytwentyfour' ); ?></p>
+<p style="font-size:0.9rem"><a href="https://tt4writer.mystagingwebsite.com"><?php echo esc_html__( 'Latest inflation report', 'twentytwentyfour' ); ?></a><?php echo esc_html__( '↗', 'twentytwentyfour' ); ?></p>
 <!-- /wp:paragraph -->
 
 <!-- wp:paragraph {"style":{"typography":{"fontSize":"0.9rem"}}} -->
-<p style="font-size:0.9rem"><a href="https://tt4writer.mystagingwebsite.com"><?php echo esc_html__( 'Applications financières pour les familles', 'twentytwentyfour' ); ?></a><?php echo esc_html__( '↗', 'twentytwentyfour' ); ?></p>
+<p style="font-size:0.9rem"><a href="https://tt4writer.mystagingwebsite.com"><?php echo esc_html__( 'Financial apps for families', 'twentytwentyfour' ); ?></a><?php echo esc_html__( '↗', 'twentytwentyfour' ); ?></p>
```
MaggieCabrera: Let's not change the content of the original text from the theme

MaggieCabrera:
I don't think these are too helpful for translators, since we are basically repeating the same thing. What is "Sample content"? I think we need to be a little more specific than that as to what the text we want translated means. Is it a description, a heading, a label for a button? What is the text talking about in the context where it's placed?

patterns/sidebar-writer.php
shail-mehta: @MaggieCabrera 
I have Added original text in theme

shail-mehta:
> I don't think these are too helpful for translators, since we are basically repeating the same thing. What is "Sample content"? I think we need to be a little more specific than that as to what the text we want translated means. Is it a description, a heading, a label for a button? What is the text talking about in the context where it's placed?

@MaggieCabrera I have changed sample title as per related context 

Please review & let me know if any corrections

shail-mehta:
@MaggieCabrera
 can you please review?

patterns/sidebar-writer.php
```diff
@@ -25,7 +25,7 @@
 <!-- /wp:separator -->
 
 <!-- wp:heading {"style":{"typography":{"fontSize":"1.6rem"}}} -->
-<h2 class="wp-block-heading" style="font-size:1.6rem"><?php echo esc_html__( 'Derniers articles', 'à propos du titre des Derniers articles', 'twentytwentyfour' ); ?></h2>
+<h2 class="wp-block-heading" style="font-size:1.6rem"><?php echo esc_html_x( 'Derniers articles', 'à propos du titre des Derniers articles', 'twentytwentyfour' ); ?></h2>
```
carolinan: This text needs to be in English.

shail-mehta: @carolinan 

its already in English in latest commit. can you please review latest commit ?
Latest Commit : https://github.com/WordPress/twentytwentyfour/pull/195/commits/851682d7b35d24a153dd903587970991b973f6e6

patterns/sidebar-writer.php
carolinan: The text **must** be in the same language in the whole theme unless the content is a quote or a name, otherwise, it becomes very difficult for translators.

Any text that is a quote in a different language than English must be labeled in the markup as being in a different language.

carolinan: (Remember that this is not demo content used to test different languages, this is placeholder content that will be used as the default)

patterns/sidebar-writer.php
carolinan: There is still several places that do not have English text.

shail-mehta: @carolinan 

Yes, I already translated in English in previous commits but @MaggieCabrera has suggested to do not change text. so I have reverted it.
https://github.com/WordPress/twentytwentyfour/pull/195#discussion_r1315702220

MaggieCabrera: the theme has 3 "variations" business (default), writer and portfolio. Each of them has the content in a different language. This is intentional and I asked @kafleg and it's something that it's ok to have on a theme in the directory, so I don't feel like there's a reason to change the content. Especially not using google translate to do so.

carolinan: Doing that would make it very difficult to translate the theme, you would need 3 times the number of translators for each supported language, or translators that speak four languages.
I disagree strongly with this. I do understand the decision is not mine to make, and I apologize for creating extra work for everyone. I was acting on what I believe is the best practice and a requirement also for WordPress core. I would like to ask you to please discuss this with the polyglots team, and that you keep discussions public.
There is no way for me to know that this has been discussed already.

MaggieCabrera: No, there hasn't been any discussion. @beafialho, do you think we could change this?

MaggieCabrera (changes requested):
I'm blocking this for the moment since we are still iterating on the patterns. Let's leave translations for when the patterns are more stable

patterns/sidebar-writer.php
aristath: WordPress Core **requires** all default text to be in US-English. 
en_us is the default language of all code in WP, and if translation files are missing, all text is considered to be the default language. Adding text in other languages will be _more difficult_ instead of easier.

afercia:
> No, there hasn't been any discussion

I would like to kindly remind everyone that WordPress is an open source project. Discussions _must_ be public, especially when they are around decisions like this one, which is unprecedented.

Technically:
I doubt providing a default theme in 3 different languages is ok. The WordPress localization tools assume the default language is US English.

Say that I'm an user form the US. I install the theme with the variation that has French content. My WordPress install is set to use US English in the front end, but the theme still displays French content. How do I switch to US English? As far as I know, there is no waty to trantlate from another language _to_ US English.

beafialho:
I had no idea this was a requirement when designing the theme, thank you for providing that information. The intent was to have three different use cases that could be from any part of the world, hence, the usage of other languages. But I agree with having all patterns be in English once they're more stable, as @MaggieCabrera suggested. We can merely _showcase_ these patterns in other languages in the demo site to maintain that intent.

MaggieCabrera:
> We can merely _showcase_ these patterns in other languages in the demo site to maintain that intent.

I think that sounds like a good idea, thank you

richtabor:
Yes, all content should all be in English out of the box. I’m not aware of discussions to change that. 

Patterns will be adjusted, but like I’ve mentioned a number of times, let’s focus on getting the design system in place - then circle back to cleaning up the sample content and adding translation functions. Otherwise iteration is much slower consistently adding them back in with every pattern change.

MaggieCabrera:
I think this PR has gone stale, now that the patterns are in a more stable state, I suggest we open new PRs to cover what needs translating/escaping

PR #204, opened 2023-09-05
Use presets for services pattern

### LINKED ISSUE #164: Services CTA pattern is almost hardcoded
**Description**

Spaces, background, font, font weight, etc of the Services CTA button are hardcoded. 

1. Go to 'patterns >> services-cta.php'
2. Check the code of the pattern one by one.

**Expected behavior**

Instead of arbitrary values, let's use the presets for the background, font, paddings, font size, font weight, etc.

richtabor:
Closes #164. 

CleanShot 2023-09-05 at 15 24 32
CleanShot 2023-09-05 at 15 24 49

### DIFF
--- a/patterns/services-cta.php
+++ b/patterns/services-cta.php
@@ -4,7 +4,6 @@
  * Slug: twentytwentyfour/services-cta
  * Categories: call-to-action
  */
-
 ?>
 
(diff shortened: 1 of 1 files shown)


richtabor:
Note I think we should hold off on 100% translatable text currently, as the patterns are still quite in flux. Once we're solid on the presets/style/width settings/etc, then we can double back and add the functions in place of raw text. Otherwise we'll keep having to spend time plugging those back every time we update.

MaggieCabrera (approved):
Thank you! I restored the translations because I didn't want to lose the descriptions for translators since that was work already done. It's fine if it's just the functions that we are missing with the changes, but I didn't want to miss that effort from other contributors

Issue #190, opened 2023-09-04
Block pattern: Full page Business About


MaggieCabrera:
Let's create a pattern for this page that makes use of all the patterns that we've already created (maybe some are not merged yet, but I think all of them have been started)

![Business About](https://github.com/WordPress/twentytwentyfour/assets/3593343/1dba020c-8c01-4f21-8220-579c70ba356e)

PR #208, opened 2023-09-05
Create about-2 pattern

richtabor:
We can adjust the nomenclature and translations later. 

CleanShot 2023-09-05 at 16 29 06
CleanShot 2023-09-05 at 16 29 23

### DIFF
--- a/patterns/about-2.php
+++ b/patterns/about-2.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * Title: About 2
+ * Slug: twentytwentyfour/about
+ * Categories: about, featured, text
+ */
+
+?>
+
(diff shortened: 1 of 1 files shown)


richtabor:
One of the patterns in https://github.com/WordPress/twentytwentyfour/issues/190.

patterns/about-2.php
```diff
@@ -0,0 +1,42 @@
+<?php
+/**
+ * Title: About 2
+ * Slug: twentytwentyfour/about
+ * Categories: about, featured, text
```
MaggieCabrera: I agree we can decide on the naming later but now we have 2 patterns with the same slug. Also, about is not a default category, is it?

MaggieCabrera (approved):
Just one comment, with that change this is good to go

patterns/about-2.php
richtabor: I'll change the slug to match the name (about-2). About is a default category; although we don't use it much (yet).

MaggieCabrera: TIL, thank you!

PR #211, opened 2023-09-05
Fix grid features pattern

richtabor:
Cleans up the feature grid pattern and uses presets throughout. 

CleanShot 2023-09-05 at 17 22 10
CleanShot 2023-09-05 at 17 22 52

### DIFF
--- a/patterns/grid-of-features.php
+++ b/patterns/grid-of-features.php
@@ -2,17 +2,18 @@
 /**
  * Title: Grid of Features
  * Slug: twentytwentyfour/grid-of-features
- * Categories: featured
+ * Categories: featured, text
  */
 ?>
-<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50"}}},"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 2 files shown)


patterns/grid-of-features.php
```diff
@@ -2,181 +2,122 @@
...
+<div class="wp-block-group"><!-- wp:paragraph {"align":"center","style":{"elements":{"link":{"color":{"text":"var:preset|color|contrast-2"}}}},"textColor":"contrast-2","fontSize":"large"} -->
+<p class="has-text-align-center has-contrast-2-color has-text-color has-link-color has-large-font-size">✳︎</p>
 <!-- /wp:paragraph -->
 
-<!-- wp:group {"style":{"spacing":{"blockGap":"0px"}},"layout":{"type":"flex","orientation":"vertical","justifyContent":"center"}} -->
-<div class="wp-block-group"><!-- wp:heading {"textAlign":"center","fontSize":"x-large"} -->
-<h2 class="wp-block-heading has-text-align-center has-x-large-font-size"><?php echo esc_html_x( 'A passion for creating spaces', 'Heading of the features', 'twentytwentyfour' ); ?></h2>
+<!-- wp:heading {"textAlign":"center","fontSize":"x-large"} -->
+<h2 class="wp-block-heading has-text-align-center has-x-large-font-size">A passion for creating spaces</h2>
```
huzaifaalmesbah: why remove `esc_html_x`

MaggieCabrera (approved):

patterns/grid-of-features.php
richtabor: Just temporary while we're ironing out patterns. We can follow up with adding these back when we're solid on patterns and no more expected changes will happen. Otherwise it slows down iterative progress.

PR #212, opened 2023-09-05
Add stretch to about-2 pattern content area

richtabor:
Before: 

CleanShot 2023-09-05 at 17 27 53

After: 

CleanShot 2023-09-05 at 17 27 37


luminuu:
Nice catch, I just added the translation strings for this pattern.

MaggieCabrera (approved):

PR #218, opened 2023-09-06
Loaded individual patterns instead of home.php pattern

kafleg:
In home.html, the entire home.php pattern was loaded and this PR will load the individual pattern of the home page.


templates/home.html
```diff
@@ -1,9 +1,16 @@
 <!-- wp:group {"style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"0","right":"0"},"blockGap":"0","margin":{"top":"0","bottom":"0"}}},"layout":{"type":"default"}} -->
-<div class="wp-block-group" style="margin-top:0;margin-bottom:0;padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:group {"style":{"position":{"type":"sticky","top":"0px"}},"layout":{"type":"default"}} -->
+<div class="wp-block-group"
+	style="margin-top:0;margin-bottom:0;padding-top:0;padding-right:0;padding-bottom:0;padding-left:0">
+	<!-- wp:group {"style":{"position":{"type":"sticky","top":"0px"}},"layout":{"type":"default"}} -->
 	<div class="wp-block-group"><!-- wp:template-part {"slug":"header-copy"} /--></div>
 	<!-- /wp:group -->
 
-<!-- wp:pattern {"slug":"twentytwentyfour/home"} /-->
```
MaggieCabrera: I don't think we should lose the home pattern yet. It can get reused for other pages the same way [this one](https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/page-business-about.php) does

kafleg: Oh, you mean loading patterns in the home.php patterns? Okay, I'll follow it and update the PR.

huzaifaalmesbah (approved):

Issue #199, opened 2023-09-05
Border radius applied to images rather than image block


richtabor:
We have a number of images within the assets directory that are using hard-cropped border radii in the image, rather than applying the border radius from Global Styles. We should have all images bundled with 0 border radius, and apply a holistic (and user-customizable) global style for the image (and other related elements) border radius. 

<img width="1218" alt="CleanShot 2023-09-05 at 15 06 47" src="https://github.com/WordPress/twentytwentyfour/assets/1813435/370678a7-45b6-421e-be5e-9990cd91e866">
<img width="685" alt="CleanShot 2023-09-05 at 15 08 13" src="https://github.com/WordPress/twentytwentyfour/assets/1813435/950b8d59-e3b2-445e-9c30-ae23503a970e">

(and others).

richtabor:
The current method also causes issues if you apply an aspect ratio.

colorful-tones:
I would be happy to tackle this across all patterns/templates once we all agree on a solution. I've added an initial PR #219 with one solution, but open to any ideas. Thanks!

richtabor:
All look good other than the services image; let's re-export that without the radii. 

<img width="929" alt="CleanShot 2023-09-07 at 16 03 05" src="https://github.com/WordPress/twentytwentyfour/assets/1813435/c6edb136-b1bd-47d0-8ec6-7b025cc4d3be">

colorful-tones:
> All look good other than the services image

Yeah, I noted that in #235:

> I found a few that had filters applied and I could not export them with edit access. We'll likely need to re-export all the images for https://github.com/WordPress/twentytwentyfour/issues/44 . So, I'm leaving it as is.

We need somebody with edit access to Figma to re-export that file with the black-and-white filter applied, but no border radius.

Or we can wait and have all images addressed in #44

MaggieCabrera:
> We need somebody with edit access to Figma to re-export that file with the black-and-white filter applied, but no border radius.
> 

Why don't we use duotone here?

colorful-tones:
#289 adds the duotone to the Services image.

PR #219, opened 2023-09-06
Try/image block border radius

colorful-tones:
**Description**

Kick off exploration for solution to #199 

I recommend two steps for each pattern/template where images are used:

- [ ] Add the `is-style-rounded` in each pattern/template as a unique commit
- [ ] Verify that each corresponding image has no radius in original image file. Re-export if necessary (requires edit access to Figma file)

I recommend using `px` for `border-radius`. One of the few properties that I would recommend using `px` for, as opposed to type and spacing. Mostly, because `border-radius` is something we want consistency with and not different calculated values for.

Note: I did not re-export the image in my example pattern, because I do not have edit access to the Figma file and one cannot disable the border-radius applied to the image within Figma when attempting to re-export, unless they have edit access.

Also, I would follow add more commits to this to cover all images and patterns if we are happy with the approach.

An alternative approach is applying the border-radius on all Image blocks and not utilize the Rounded variation. Example of `theme.json`:

```
			"core/image": {
				"border": {
					"radius": "16px"
				}
			},
```

**Screenshots**

Screenshot 2023-09-06 at 9 59 39 AM

**Testing Instructions**

Check out the 'Project description' pattern and `theme.json` entry for `core/image` block

### DIFF
--- a/patterns/about-2.php
+++ b/patterns/about-2.php
@@ -35,8 +35,8 @@
 <!-- /wp:column -->
 
 <!-- wp:column {"verticalAlignment":"center","width":"50%"} -->
-<div class="wp-block-column is-vertically-aligned-center" style="flex-basis:50%"><!-- wp:image {"aspectRatio":"3/4","scale":"cover","sizeSlug":"large","linkDestination":"none"} -->
-<figure class="wp-block-image size-large"><img src="<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/tt4_about_1.jpg" alt="<?php esc_attr_e( 'Picture Of Business About', 'twentytwentyfour' ); ?>" style="aspect-ratio:3/4;object-fit:cover"/></figure>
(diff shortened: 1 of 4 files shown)


richtabor:
> I recommend using px for border-radius. One of the few properties that I would recommend using px for, as opposed to type and spacing. Mostly, because border-radius is something we want consistency with and not different calculated values for.

We could try min/max here potentially — or even one of the fluid font sizes — if we wanted to have the radius be a bit smaller on mobile. Worth a try imo.

richtabor:
I think this could work. 

The tricky thing we _may_ need to iron out is adding a rounded style to post featured image and quote, as those also utilize the border-radius style.

richtabor (changes requested):
This could work. The idea here is that we use the "Rounded" block style for image blocks; instead of forcing a default border radius to all images. 

I updated a couple patterns to use is-style-rounded, but we'd need to make sure we're not hard-coding a 16px radius to images throughout all patterns (to use this approach consistently).

colorful-tones:
Replaced with #235

PR #182, opened 2023-09-02
Add centered call to action

carolinan:
**Description**

This PR replaces https://github.com/WordPress/twentytwentyfour/pull/43 and closes https://github.com/WordPress/twentytwentyfour/issues/114

**Screenshots**

Call to action with heading, paragraph, and button

**Testing Instructions**
Open the pattern inserter and select the "call to action" category. Locate and insert the pattern.

**Contributors**
Props @LynUrth

PR #220, opened 2023-09-06
Created Centered call to Action pattern

kafleg:
This pattern is copied from home.php pattern to make a new pattern.


huzaifaalmesbah:
already Merged #182

kafleg:
Thanks for checking. I didn't find it before as the file name was just cta.php 

I've updated my another PR. 

Closing this PR.

PR #240, opened 2023-09-07
Fix default separator width issue

### LINKED ISSUE #239: Separator Block default and wide line styling outputing same result.
### Screenshots: 

**When Default is selected** : https://prnt.sc/rP-MIZEnnZjV
**When Wide is selected**    : https://prnt.sc/nio2ni4mT_dR

shail-mehta:
Closes https://github.com/WordPress/twentytwentyfour/issues/239.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -656,7 +656,8 @@
 				},
 				"color": {
 					"text": "var(--wp--preset--color--contrast)"
-				}
+				},
+				"css": " &:not(.is-style-wide):not(.is-style-dots):not(.alignwide):not(.alignfull){width: 200px}"
 			},
 			"core/site-tagline": {
 				"typography": {
--- a/theme.json
+++ b/theme.json
@@ -657,7 +657,7 @@
 				"color": {
 					"text": "var(--wp--preset--color--contrast)"
 				},
-				"css": " &:not(.is-style-wide):not(.is-style-dots):not(.alignwide):not(.alignfull){width: 100px}"
+				"css": " &:not(.is-style-wide):not(.is-style-dots):not(.alignwide):not(.alignfull){width: var(--wp--preset--spacing--60)}"
 			},
 			"core/site-tagline": {
 				"typography": {
--- a/theme.json
+++ b/theme.json
@@ -657,7 +657,7 @@
 				"color": {
 					"text": "var(--wp--preset--color--contrast)"
 				},
-				"css": " &:not(.is-style-wide):not(.is-style-dots):not(.alignwide):not(.alignfull){width: 200px}"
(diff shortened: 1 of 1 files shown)


theme.json
```diff
@@ -656,7 +656,8 @@
 				},
 				"color": {
 					"text": "var(--wp--preset--color--contrast)"
-				}
+				},
+				"css": " &:not(.is-style-wide):not(.is-style-dots):not(.alignwide):not(.alignfull){width: 100px}"
```
MaggieCabrera: I'd rather not include css like this when we can actually target the separator block variation [like this](https://github.com/WordPress/twentytwentyfour/blob/cd4565a27f0a48b1a94bd9307fe4ea2e97a61bdf/theme.json#L458)

MaggieCabrera (dismissed):

theme.json
richtabor: So, while we can target the `default` variation, `css` does not look like it renders for each variation. 

This for example, does not work: 

CleanShot 2023-09-07 at 09 16 50

While this does: 

CleanShot 2023-09-07 at 09 17 51

carolinan: Correct, CSS is not supported for elements or variations.
https://github.com/WordPress/gutenberg/issues/53186

carolinan:
What about unregistering the wide style and just using the alignment setting on the block?

theme.json
richtabor: > CSS is not supported for elements or variations.

I think it could be cool for variations.

carolinan: All our wishes will not be fulfilled in time for 6.4 😄 
For example, it needs a UI, otherwise, users can not manage the custom CSS.

richtabor:
> What about unregistering the wide style and just using the alignment setting on the block?

I don't think it's consistent enough. We still have oddities with width when items are within a stack block for example. Eventually it should just support a width setting though. 

I don't mind the current approach positioned here (same as Twenty Twenty-Three) though.

carolinan:
Yeah the stack issue has been known for some time without a solution :(

theme.json
richtabor: I wouldn't think that's a blocker—but alas, I'm not thinking custom variation css will land soon.

richtabor:
> Yeah the stack issue has been known for some time without a solution :(

Yea, it's very rough; essentially makes the Stack block quite less useful/capable (if you want to size spacers, or others, within it).  Do you recall the issue on Gutenberg?

richtabor (approved):
LGTM. In the future perhaps the block could have a proper width control, but for now this works fine. I also used the fluid spacing preset as the width, so that there's some mobile fluidity to it. 

CleanShot 2023-09-07 at 16 37 39
CleanShot 2023-09-07 at 16 42 40

carolinan (approved):

PR #261, opened 2023-09-07
Templates: Home cleanup

MaggieCabrera:
**Description**

This PR moves the group wrapper inside the pattern, fixes the meta for the pattern so it can be reused and adds the main tag on the wrapper for a11y.


huzaifaalmesbah (commented):
Wow. looks good.

huzaifaalmesbah (approved):

PR #263, opened 2023-09-07
Add/additional font size

pbking:
**Description**

This change adds an additional font size.  (XX-Large)

However, since a SMALLER font size is what is needed this shifts ALL of the font sizes UP

The value of 'small' becomes the value of 'medium' and is assigned as the default size.
The value of 'medium' becomes the value of 'large'
The value of 'large' becomes the value of 'x-large'
The value of 'x-large' becomes the value of the NEW size 'xx-large'

And SMALL becomes a NEW value of `clamp(0.813rem, 0.688rem)`
(I'm not tied to this value, it just seemed to hit most of what was needed based on the explicit values used in existing patterns)

Additionally the font size values have been added to the readme.md file for easy reference.

**Screenshots**

There are no visible changes to the theme, patterns, layout, etc (assuming I made all of the changes correctly...)

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -294,7 +294,7 @@
 						"max": "3rem",
 						"min": "2.5rem"
 					},
-					"name": "Extra Large",
+					"name": "Extra Extra Large",
 					"size": "3rem",
 					"slug": "xx-large"
 				}
--- a/theme.json
+++ b/theme.json
@@ -358,7 +358,7 @@
 					"blockGap": "0.7rem"
 				},
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--small)"
+					"fontSize": "var(--wp--preset--font-size--medium)"
 				}
 			},
 			"core/code": {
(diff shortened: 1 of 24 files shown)


theme.json
```diff
@@ -253,41 +253,50 @@
 				}
 			],
 			"fontSizes": [
+				{
+					"fluid": {
+						"max": "0.813rem",
+						"min": "0.688rem"
```
MaggieCabrera, line 259: I would keep this one non fluid since it's small as it is. Just let it be 0.8 on any screen size

MaggieCabrera:
This is what I would have done too, thank you @pbking! I would love feedback on this from both design and a11y perspective. I agree the new font size is a bit too small for comfort, but I think if we keep it on just a few use cases like the post meta, it should be fine?

richtabor:
I'm looking now.

pbking:
Closing in lieu of #268

PR #268, opened 2023-09-07
Add fluid type scale

richtabor:
**Description**

An alternate (and inspired by) to https://github.com/WordPress/twentytwentyfour/pull/263. This only changes the theme.json; not patterns, as we'll have conflicts doing both together while patterns are changing yet. 

Key points: 
- Uses a 1.333 ratio, starting at 1.05rem (effectively 17px) as a body font size. 
- The "medium" font size is used as the body/default font size. 
- Headings now use font size presets. 
- Small and Medium font sizes are not fluid (to ensure readability on mobile). 
- Large, XL, and XXL font sizes scale down one type-scale size lower on mobile. 

Others: 
- Updated button to use the type scale (and removed duplicate button styling). 

### Visuals
Desktop: 
CleanShot 2023-09-07 at 11 28 06

Mobile: 
CleanShot 2023-09-07 at 11 29 59

### Patterns
We do need to follow-up with changes to a few patterns; but generally they flow pretty well. Here's a few shots of patterns that don't need changes: 

CleanShot 2023-09-07 at 11 33 02
CleanShot 2023-09-07 at 11 33 58
CleanShot 2023-09-07 at 11 35 44
CleanShot 2023-09-07 at 11 33 50
CleanShot 2023-09-07 at 11 36 15

PR #271, opened 2023-09-07
Bring back translations to the pricing pattern

juanfra:
**Description**

- Bring back translations to the pricing pattern.


richtabor:
Solid, but do want to note that there are still changes that may land. We're starting to get solid on the design system (esp if/when fluid type scale https://github.com/WordPress/twentytwentyfour/pull/268 lands). Mind taking a look at that?

carolinan:
I honestly don't understand why changes to the design system keeps removing the translations... 
Example: Changing a size preset on a wrapping group block in a pattern should not change any text at all, only the small part of the pattern that *needs* to be changed, should be changed.

richtabor:
> I honestly don't understand why changes to the design system keeps removing the translations...

Because we're not just changing one or two values; they are more substantial. So we're having to copy/paste the new pattern code into the PHP file, then add back all the translation functions. It's quite time-consuming and slows down the iterative process. 

About half of the patterns are solid; will less possible changes — but the other half still need a good bit of re-working. If you look at the patterns in the library, you can tell them apart well.

juanfra:
> Solid, but do want to note that there are still changes that may land. We're starting to get solid on the design system (esp if/when fluid type scale https://github.com/WordPress/twentytwentyfour/pull/268 lands). Mind taking a look at that?

Sure, will do 🎉

carolinan (approved):

PR #274, opened 2023-09-07
Add offset image query loop pattern

scruffian:
**Description**
This is an alternative to https://github.com/WordPress/twentytwentyfour/pull/136

This turned out to be quite tricky to achieve. To me the design looks like a masonry pattern, so I went with one columns block with 4 columns inside it, and each of the columns containing one query loop each with 3 items. This means that the order reads from top to bottom rather than left to right, which may not be the intention.

I tried for a long time to use a margin on the Post Featured Image block, which works fine, except when the columns collapse the margins either add to each other and leave uneven space, or there is no margin. In the end I changed the approach to use block spacing inside the post template and then added spacer blocks to offset the two even columns. This means that there is no bottom/top margin to get in the way.

**Screenshots**
Desktop:
Screenshot 2023-09-07 at 17 35 34

Mobile:
Screenshot 2023-09-07 at 17 35 51

**Testing Instructions**
Add the offset pattern to a post or check the home-portfolio template.


richtabor (approved):
This works for me! Added a few tweaks to make sure it fits right as a pattern within a page pattern, as well as standalone. 

https://github.com/WordPress/twentytwentyfour/assets/1813435/dc69f16f-2f5c-4a80-846b-ce81e9a00833

jordesign:
Just wanted to say this looks great - and I love the approach taken ❤️

PR #275, opened 2023-09-07
Update posts pattern

### LINKED ISSUE #262: Patterns: Posts pattern needs to have clickable featured images and post titles
This pattern is used in the home so we need to have the ability to navigate to the posts pages via clicking their titles or featured images.

richtabor:
**Description**
Closes #262, adds spacing presets, and utilizes aspect ratio on featured images to have the pattern site better. 

**Screenshots**

CleanShot 2023-09-07 at 12 44 28

### DIFF
--- a/patterns/posts.php
+++ b/patterns/posts.php
@@ -74,9 +74,6 @@
 
 <!-- wp:post-author-name {"isLink":true} /-->
 
-<!-- wp:paragraph -->
-<p>—</p>
-<!-- /wp:paragraph -->
 
 <!-- wp:post-terms {"term":"category"} /--></div>
 <!-- /wp:group --></div>
--- a/patterns/posts.php
+++ b/patterns/posts.php
@@ -7,13 +7,13 @@
  */
 ?>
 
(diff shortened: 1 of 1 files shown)


huzaifaalmesbah (commented):
I tested, and now title or featured image is clickable. good work.

patterns/posts.php
```diff
@@ -6,55 +6,80 @@
...
+<!-- wp:paragraph -->
+<p>—</p>
 <!-- /wp:paragraph -->
 
-<!-- wp:post-author-name {"isLink":true,"style":{"elements":{"link":{"color":{"text":"var:preset|color|contrast-2"}}}},"textColor":"contrast-2"} /-->
+<!-- wp:post-author-name {"isLink":true} /-->
+
+<!-- wp:paragraph -->
+<p>—</p>
+<!-- /wp:paragraph -->
```
MaggieCabrera:
```suggestion
```

MaggieCabrera (approved):
I'm approving this but I would like @beafialho to have a look. We didn't separate this into 3 query loops, but the design suggest that we should because all 3 images have different aspect ratio. I'm not in love with the size difference of the images, but it's not my call to make

beafialho:
I'd like to increase the post title size on the left and have a bit less spacing between the two posts on the right. Other than that, this looks good, thanks!

https://github.com/WordPress/twentytwentyfour/assets/65220155/c9daad67-7fed-4f05-ae2c-001ec2533232

PR #279, opened 2023-09-07
Use presets for outer spacing

### LINKED ISSUE #224: Mobile issues
Let's make sure that all templates and patterns look fine on mobile. We don't have mobile designs for every pattern, but we can use these as guidelines as to how the theme is expected to look:

Screenshot 2023-09-06 at 18 25 12

There must be something wrong with the post template because I see no padding when I insert a pattern that does look fine in the home page:

richtabor:
Uses the spacing presets for outer spacing; making it fit a bit better for mobile + matching the spacing used in patterns. 

Closes #224. 

### Visual: 

CleanShot 2023-09-07 at 15 53 26

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -785,10 +785,8 @@
 		"spacing": {
 			"blockGap": "1.2rem",
 			"padding": {
-				"bottom": "0",
 				"left": "var(--wp--preset--spacing--50)",
-				"right": "var(--wp--preset--spacing--50)",
-				"top": "0"
+				"right": "var(--wp--preset--spacing--50)"
 			}
 		},
 		"typography": {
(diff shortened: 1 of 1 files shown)


theme.json
```diff
@@ -785,10 +785,10 @@
 		"spacing": {
 			"blockGap": "1.2rem",
 			"padding": {
-				"bottom": "0vh",
-				"left": "6vw",
-				"right": "6vw",
-				"top": "0vh"
+				"bottom": "0",
```
MaggieCabrera: it's better if we don't define as 0 the vertical padding, else we have a [bug with the nav menu](https://github.com/WordPress/gutenberg/pull/53725)

MaggieCabrera (approved):

PR #283, opened 2023-09-07
Add Quote block global styling

### LINKED ISSUE #257: Quote default and plain variations are the same
We need to implement the quote block's styling within global styles, also accounting for the variations. 

Current: 

CleanShot 2023-09-07 at 09 21 35
CleanShot 2023-09-07 at 09 21 30

Expected: 

I would say "the first one should be "Plain" and the white background quote should be "Default".
CleanShot 2023-09-07 at 09 22 24

colorful-tones:
**Description**

Addresses #257 

**Screenshots**

Screenshot 2023-09-08 at 10 33 57 AM

~~Note: I have to do a `"css"` entry to try and override the default `core/paragraph` font-size styling. However, it might be recommended to remove the `fontSize` on `core/paragraph`. This should just inherit the `body` `font-size`.~~ 🤔 
~~Note: this required the removal of the existing Paragraph block `"core/paragraph"` font styling, which should mostly be inherited from the `body` elements typography entry.~~

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -597,17 +597,55 @@
 			},
 			"core/quote": {
 				"border": {
-					"color": "var(--wp--preset--color--primary)",
-					"style": "solid",
-					"width": "0 0 0 1px"
+					"radius": "var(--wp--preset--spacing--20)"
+				},
+				"color": {
+					"background": "var(--wp--preset--color--base-2)"
+				},
+				"css": "&:not(.is-style-plain) :where(p) {font-size:var(--wp--preset--font-size--large);} &.is-style-plain :where(cite) {font-size:var(--wp--preset--font-size--small);} &.is-style-plain :where(p) {font-size:var(--wp--preset--font-size--medium);} :where(p) {margin-block:0;}",
+				"elements": {
+					"cite": {
+						"typography": {
+							"fontFamily": "var(--wp--preset--font-family--system-font)",
+							"fontSize": "var(--wp--preset--font-size--medium)",
+							"fontStyle": "normal",
+							"lineHeight": "1"
+						}
+					}
 				},
 				"spacing": {
 					"padding": {
-						"left": "var(--wp--preset--spacing--50)"
+						"bottom": "var(--wp--preset--spacing--30)",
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
> Note: I have to do a `"css"` entry to try and override the default `core/paragraph` font-size styling. However, it might be recommended to remove the `fontSize` on `core/paragraph`. This should just inherit the `body` `font-size`. 🤔

I think that's what we should do too.

richtabor:
> I think that's what we should do too.

I removed it as part of #268.

colorful-tones:
I've removed the `"core/paragraph"` font styling and streamlined the `"css"` and updated the screenshot in original description (at top).

theme.json
```diff
@@ -489,13 +489,6 @@
 					"fontWeight": "500"
 				}
 			},
-			"core/paragraph": {
-				"typography": {
-					"fontSize": "var(--wp--preset--font-size--small)",
-					"letterSpacing": "0.01em",
-					"lineHeight": "1.5"
-				}
-			},
```
richtabor: This shouldn't be necessary. Can you rebase to pull in today's changes?

colorful-tones: Good call, and rebased. Ty!

theme.json
```diff
@@ -597,17 +597,58 @@
...
 				"border": {
-					"color": "var(--wp--preset--color--primary)",
-					"style": "solid",
-					"width": "0 0 0 1px"
+					"radius": "var(--wp--preset--spacing--20)"
+				},
+				"color": {
+					"background": "var(--wp--preset--color--base-2)"
+				},
+				"css": "& :where(p) {margin-block-start:0;margin-block-end:var(--wp--preset--spacing--10);} &.is-style-plain :where(cite) {font-size:var(--wp--preset--font-size--small);}",
```
richtabor: > & :where(p) {margin-block-start:0;

That should probably be resolved in core eventually. Not sure if there's an issue yet.

richtabor (approved):
LGTM. 

CleanShot 2023-09-11 at 12 19 13

PR #294, opened 2023-09-08
Home portfolio template

MaggieCabrera:
**Description**

Cleans up the home portfolio template and reworks the footer for the portfolio related templates too. Strings pending on any changes. I'm not sure what to do about the heading in the template, if we should move it to the pattern for translation or if we should turn it into using page title block. that won't work if the user replaces their home page with this template, but we don't have a flow for replacement yet.

The Offset Grid of Image Posts pattern needs a bit of work, it doesn't look great on my local version if all my images have the same aspect ratio. Should we force the aspect ratio on the featured images to make sure the design is similar to the original idea?

/cc @beafialho @richtabor


richtabor:
> The Offset Grid of Image Posts pattern needs a bit of work, it doesn't look great on my local version if all my images have the same aspect ratio. Should we force the aspect ratio on the featured images to make sure the design is similar to the original idea?

I don't think so. Artists are very particular about how images display (almost always as taken/intended—not forced to a specific aspect ratio). I think it works with images that are all the same, and with images of all shapes.

richtabor:
I pushed a few changes to make it a bit more portable, where we can use it as a starter page pattern, as well as within the template. 

I think this concept will work for all the extra templates, where headers and footers are in the templates, and patterns are used for the contents — which are also used as a page pattern. 

CleanShot 2023-09-08 at 18 10 33
CleanShot 2023-09-08 at 18 10 52

richtabor:
The headers/footers need a bit of tidying up too (presets, using spacer blocks instead of padding where it makes sense), but I'm not too concerned about those atm.

richtabor:
> if we should move it to the pattern for translation or if we should turn it into using page title block.

Generally keeping them localized in patterns is a good approach. Also, we should be using paragraph blocks for lengthy text like these (but with the font, size and line height (1.2) used in headings. I've converted a few, but there's probably more as we dig in.

richtabor (approved):
LGTM. We'll keep iterating on the footer/headers, but generally let's lean in on this direction and explore how it comes together.

PR #308, opened 2023-09-10
Some missing translations enable

kamrulhasanj:
**Description**
Some missing translations are enabled in TT4

### DIFF
--- a/patterns/footer-portfolio.php
+++ b/patterns/footer-portfolio.php
@@ -19,18 +19,18 @@
 <!-- wp:columns {"style":{"spacing":{"padding":{"top":"2vh"}}}} -->
 <div class="wp-block-columns" style="padding-top:2vh"><!-- wp:column {"width":"57%"} -->
 <div class="wp-block-column" style="flex-basis:57%"><!-- wp:paragraph {"fontSize":"x-large","fontFamily":"cardo"} -->
-<p class="has-cardo-font-family has-x-large-font-size">Resta al passo, contattaci.</p>
+<p class="has-cardo-font-family has-x-large-font-size"><?php echo esc_html_x( 'Resta al passo, contattaci.', 'Footer Portfolio left paragraph', 'twentytwentyfour' ); ?></p>
 <!-- /wp:paragraph --></div>
 <!-- /wp:column -->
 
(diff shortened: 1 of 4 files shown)


patterns/footer-portfolio.php
```diff
@@ -58,7 +58,7 @@
 
 <!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
 <div class="wp-block-group"><!-- wp:paragraph {"fontSize":"small"} -->
-<p class="has-small-font-size">© 2024 Twenty Twenty Four</p>
+<p class="has-small-font-size"><?php echo esc_html_x( '© 2024 Twenty Twenty Four', '', 'twentytwentyfour' ); ?></p>
```
huzaifaalmesbah, line 61: use like that `© 2024 Twenty Twenty-Four` and give context

patterns/footer-portfolio.php
```diff
@@ -58,7 +58,7 @@
...
 <!-- /wp:paragraph -->
 
 <!-- wp:paragraph {"fontSize":"small"} -->
```
huzaifaalmesbah, line 64: You also follow the footer.php, and enable the translate 
`<p class="has-small-font-size">Designed with <a rel="nofollow" href="https://wordpress.org">WordPress</a></p>`

MaggieCabrera (changes requested):
Let's hold off with the translations until all the patterns have been reviewed

patterns/footer-portfolio.php
juanfra: Twenty Twenty Four is probably something we don't want to translate. This string should be processed so the copyright year comes from a variable and not part of the translatable string.

carolinan: I think this should be the site title block.

patterns/footer-portfolio.php
juanfra: Please check these lines: https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/footer.php#L16-L24

MaggieCabrera:
Closed in favor of https://github.com/WordPress/twentytwentyfour/pull/671

PR #313, opened 2023-09-11
Fix PHP issue in the sidebar template part

carolinan:
**Description**

There is a problem with the code for the post date block in the Latest Posts section of the sidebar template part.
This prevents the single with sidebar template from displaying on the front.

The PR solves the typo in the markup.

**Testing Instructions**

Create a new post. 
Assign it the "Single with sidebar" template.
Save and view the post on the front.


juanfra (approved):
LGTM

PR #314, opened 2023-09-11
Remove image ID from patterns

juanfra:
**Description**

Remove image ID from patterns.


carolinan:
I think both of these images are supposed to be the avatar block.

MaggieCabrera:
> I think both of these images are supposed to be the avatar block.

Yeah, I think so too

juanfra:
Awesome, I'll update the patterns to use that block then.

MaggieCabrera:
Closing this in favor of https://github.com/WordPress/twentytwentyfour/pull/381

PR #316, opened 2023-09-11
Header template part cleanup

### LINKED ISSUE #56: Site Logo is paragraph instead of image
## Issue Summary

This an accessibility issue and a usability issue. The design calls for an icon to display to the left of the site title in the header block. The issue from an accessibility point of view is that the icon is decorative but it's being read out by the screen reader because ~it's in a paragraph~ it's a glyph within the readable space of unicode (see more in @joedolson's comment below). It conveys no relevant information for the user. 

From a usability perspective, this type of graphic in the design is often where a company logo would go and it's likely that many users of this theme will want to use an image here.  Having it as a paragraph doesn't allow users to easily upload their own logo to this header. 

## Current code
```HTML
<div class="wp-block-group"><!-- wp:paragraph {"style":{"typography":{"fontSize":"1.5rem"}}} -->
<p style="font-size:1.5rem">✦</p>
<!-- /wp:paragraph -->
<!-- wp:group {"style":{"spacing":{"blockGap":"0px"}}} -->
<div class="wp-block-group"><!-- wp:site-title /--></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```
## Steps to reproduce

1. Go to any page in the theme. 
2. Inspect element on the icon next to the site title in the header block of the page. 
3. Observe that it is a paragraph and not an image. 

## Recommendation
1. Swap out the paragraph block for an image block. 
2. Ensure that the alt is empty since in our demo template, the diamond is decorative `alt=""`
3. Adjust the styles of the theme so it follows the figma spec. 

### HTML
```HTML
<div class="wp-block-group"><!-- wp:image {"id":18,"width":51,"height":51,"sizeSlug":"full","linkDestination":"none"} -->
<figure class="wp-block-image size-full is-resized"><img src="logo.png" alt="" class="wp-image-18" style="width:51px;height:51px" width="51" height="51"/></figure>
<!-- /wp:image -->

<!-- wp:group {"style":{"spacing":{"blockGap":"0px"}}} -->
<div class="wp-block-group"><!-- wp:site-title /--></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```

## Specifications
**Relevant Template Files:** 
[header-copy.html](https://github.com/WordPress/twentytwentyfour/blob/trunk/parts/header-copy.html)
[header-portfolio.html](https://github.com/WordPress/twentytwentyfour/blob/trunk/parts/header-portfolio.html)
[header-writer.html](https://github.com/WordPress/twentytwentyfour/blob/trunk/parts/header-writer.html)
[header.html](https://github.com/WordPress/twentytwentyfour/blob/trunk/parts/header.html)
**WCAG Principle:** Understanding
**WCAG SC:** 1.1.1

### LINKED ISSUE #311: Header: Should header be sticky on all pages?
Currently it appears that the header is only Sticky on the Blog Home Template. Is this something that should consistent across the whole site (or at least the other Home Page variants)?

MaggieCabrera:
**Description**

This PR removes all extra headers and reduces it to just one. I've replaced the glyph with the site logo.

Closes https://github.com/WordPress/twentytwentyfour/issues/56
Closes https://github.com/WordPress/twentytwentyfour/issues/311

**Screenshots**

Screenshot 2023-09-11 at 12 14 22

@beafialho I have seen that there's just one header in the figma, but the old demo sites were using different ones to display the different glyphs, I think I made the correct assumption removing the extra template parts and keeping just one. Some places the header is sticky, some places it isn't. The same with the border bottom. We can express this by making changes to the templates themselves if needed, I just need to know when we want either of those to happen, can you let me know? I will update this PR with that info.


beafialho:
@MaggieCabrera let's make the header sticky everywhere, and let's not use the border anywhere.

One minor adjustment I ask you to do is change the a background color of the header group to `#FAFAFAF5 ` instead of `#FFFFFF`, which gives it the same color as the site background but with a nice transparency. Do you think we should add it to the color palette, for this to work well with the Style Variations?

Finally, is it possible to specify a maximum dimension to the Site Logo?

MaggieCabrera:
> One minor adjustment I ask you to do is change the a background color of the header group to `#FAFAFAF5 ` instead of `#FFFFFF`, which gives it the same color as the site background but with a nice transparency. Do you think we should add it to the color palette, for this to work well with the Style Variations?
> 

Where else would we be using that color? And what would be the alternative for that color in those style variations?

> Finally, is it possible to specify a maximum dimension to the Site Logo?

I think we can define the width only

beafialho:
> Where else would we be using that color? And what would be the alternative for that color in those style variations?

Nothing I remember. In the style variations, the shade could be `#212121ED`. Here's the expected view:

Captura de ecrã 2023-09-11, às 11 51 45

Unless we could set the respective background colors but specify the opacity separately or elsewhere?

> I think we can define the width only

Hmm, can we define the aspect ratio? I'd rather not specify width because for squared logos, a small width would work well but for long logos, it would not 🤔

MaggieCabrera:
Ok, I'll add it to the theme.json, else I don't think we can make it work with the variations.

And no, sadly the site-logo doesn't support aspect ratio that I know of

MaggieCabrera:
https://github.com/WordPress/gutenberg/issues/52169 the issue for aspect ratio on site logo

beafialho:
Ok, let's leave the site logo as is then. Thank you!

MaggieCabrera:
The "stickyness" wasn't working when applied inside the template part, so I used a wrapper and applied the header tag to it instead. I also added the extra color, let me know if you think the naming convention is not correct @richtabor 

/cc @carolinan for markup review too

richtabor:
@beafialho I think we're going to get tagged via accessibility with the transparent header background. I suggest we revert it back to the base color.

beafialho:
I think it could be worth giving it a try and revert if it's not accepted 🙂 the transparency is so small, but the detail really makes a difference.

MaggieCabrera:
@joedolson can you chime in on this, please? I'm happy to revert if needed

joedolson:
Transparency isn't fundamentally a problem; it becomes a problem if it creates a contrast issue that makes the foreground content (the header) unreadable when the background is behind it. The transparency here is subtle enough that I don't see a problem.

I am noticing that the header has two `header` elements, however, and that's a problem; it causes screen readers to double announce the banner landmark region. There should only be one `header` element; the other element should just be a `div`.

richtabor:
> Transparency isn't fundamentally a problem; it becomes a problem if it creates a contrast issue that makes the foreground content (the header) unreadable when the background is behind it. 

Agreed — but I'm not entirely confident in these results: 

CleanShot 2023-09-12 at 17 30 06
CleanShot 2023-09-12 at 17 30 39
CleanShot 2023-09-12 at 17 31 16

If we added backdrop blurring, that would help, but I don't think we should add that level of custom CSS— which can't be manipulated outside of code.

joedolson:
There aren't any problems explicitly in those examples. That's not to say it's ideal - adding extra lines and angles, however subtle, behind written text is going to make it harder to read for somebody. However, with these colors, it doesn't cause a specific WCAG violation.

If a user changed the colors, however, so that the header text was a gray that just barely passed accessibility against white, then it would become a problem.

richtabor:
Thanks @joedolson.

@beafialho I think we should play it safe and not do the transparent header. I agree it's a nice touch, though it's too fragile where folks will end up with a less than ideal result.

richtabor:
>  I think I made the correct assumption removing the extra template parts and keeping just one.

Just thinking aloud here: 

If we don't include the template parts for each, we will be unable to place different header/footer parts in different templates. 

But maybe that's not a bad thing. 

If you replace a template with another, and you get a different header, your whole site will have header A—while that template has header B. Then you have to figure out how to edit both. If there's one header template part used throughout, you can replace them with another and it happens site-wide.

beafialho:
> Thanks @joedolson.
> 
> @beafialho I think we should play it safe and not do the transparent header. I agree it's a nice touch, though it's too fragile where folks will end up with a less than ideal result.

Ok, sounds good.

MaggieCabrera:
> I am noticing that the header has two `header` elements, however, and that's a problem; it causes screen readers to double announce the banner landmark region. There should only be one `header` element; the other element should just be a `div`.

I think this has been fixed, where do you see it?

joedolson:
In the rendered source, with this PR applied: 

image
Image shows nested headers: `.wp-block-group .is-position-sticky` and `.wp-block-template-part`.

richtabor:
> Image shows nested headers: .wp-block-group .is-position-sticky and .wp-block-template-part.

Yes, this needs to be resolved.

richtabor (approved):
I think this is a decent first pass. May end up adjusting it a bit further, but let's get this in before too many conflicts.

MaggieCabrera:
Mmmh, I didn't know the header template part defaulted to using the header tag, my intention was to have the sticky wrapper be the one with the header tag, so it's first in the DOM, but I'm not sure how important that is. If need be, we can iterate on that.

richtabor:
> Mmmh, I didn't know the header template part defaulted to using the header tag

Yea, that's what I did the opposite first. It's important to keep them "header" parts, for the replace flow and organization in the UI.

MaggieCabrera:
> > Mmmh, I didn't know the header template part defaulted to using the header tag
> 
> Yea, that's what I did the opposite first. It's important to keep them "header" parts, for the replace flow and organization in the UI.

very right! thanks for pointing that out

PR #317, opened 2023-09-11
Add RSVP pattern

### LINKED ISSUE #122: Block Patterns - RSVP
**Description**

🛠️ The goal of this issue is to track the implementation of the RSVP pattern.
🎨 [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=565%3A10707&mode=design&t=LHSPHKUrPq556l59-1)

RSVP

carolinan:
**Description**

Adds the RSVP pattern
Enables writingMode in theme.json

Closes https://github.com/WordPress/twentytwentyfour/issues/122

**Screenshots**

**Testing Instructions**

**This PR requires Gutenberg to be active**

Create a new post or page.
Open the pattern inserter and add the "RSVP" pattern from the Call to Action category.
Save.
View the pattern in the editor and front and on different browser widths.


MaggieCabrera:
aaah, I'm glad that you get to be the one to make this one ;)

carolinan:
I tried both columns and the media and text block. Both have some side effects:

1. With the columns, I could not make the image fill the height of the text content. This can be done with the media and text block.
2. The media and text block has some strange default spacing on the text content part, so it looks like the spacing is larger than the padding spacing preset that I added to the wrapper. This still needs some work.

Both methods look ugly on mobile and need work.

The text orientation option does not support headings yet so for now, it is a paragraph.

MaggieCabrera:
Are there any changes to the design that could help?

MaggieCabrera:
Also the template [About](https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/about-2.php) is very very similar, does it have the same problems?

richtabor:
> I could not make the image fill the height of the text content. 

I wouldn't worry about that, but rather make the image set so that whichever size it becomes, it still looks good.

Issue #319, opened 2023-09-11
Update single.html to match Figma


jordesign:
There are some differences in the styling and spacing of elements between Figma and the existing single.html template that could be tidied up.

jordesign:
Figma:

<img width="489" alt="Screenshot 2023-09-11 at 9 03 57 pm" src="https://github.com/WordPress/twentytwentyfour/assets/1842363/c728da20-4586-4850-ad38-2114fc23c4d6">

Actual:

![theme-building local_3-years-at-automattic_](https://github.com/WordPress/twentytwentyfour/assets/1842363/7a9ab3e2-9add-47e7-920c-e2c24e09b16f)

MaggieCabrera:
I think those were tweaks that @beafialho made to the template while she was building so I'd let her review first before making any changes.

jordesign:
@MaggieCabrera @MaggieCabrera Sounds good - #320 is a PR with potential changes if they're needed 👍

beafialho:
I don't think I touched this template, or if I did it wasn't in great detail. Feel free to adjust accordingly to match with the Figma design.

PR #320, opened 2023-09-11
Updates to Single Post spacing and styling

jordesign:
**Description**

Addressing #319 - this aims to update the single.html file to be closer to the [Figma design](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=492-1457&mode=design&t=1ksmbRV1pC17uorE-0). Most notable additions.

* Update spacing between main elements (Title, Body, Comments).
* Add separator at bottom of post.
* Add CSS styling for the taxonomy/tags

**Screenshots**

theme-building local_3-years-at-automattic_ (1)

** Questions **

I did my best to use theme presets where possible but there were a few aspects I needed to clarify..

* Comments title sizing was different - but uses the closest preset sizing to what we have.
* There's no preset colors with the subtlety used for the tags/taxonomy background - so I've hard-coded that into the CSS (which I imagine we need to amend).
* Similarly - the other values in the CSS for the tags didn't have matching presets

### DIFF
--- a/style.css
+++ b/style.css
@@ -52,3 +52,20 @@ a {
 .is-style-arrow-icon-details[open] > summary {
 	list-style-type: '\2192\00a0\00a0\00a0';
 }
+
+/*
+ * Styles for the taxonomy/tags on single posts
+ * https://github.com/WordPress/twentytwentyfour/issues/XX
+ */
+
+.wp-block-post-terms a {
+	display:inline-block;
+	background-color: #f2f2f2;
+	padding: 6px 14px;
+	border-radius:16px;
+	margin: 0 10px 10px 0;
+}
+
+.wp-block-post-terms a:hover {
+	background:#eee;
+}
--- a/style.css
+++ b/style.css
@@ -55,12 +55,11 @@ a {
 
 /*
  * Styles variation for post terms
- * https://github.com/WordPress/twentytwentyfour/issues/319
+ * https://github.com/WordPress/gutenberg/issues/24956
  */
 
-.is-style-solid a,
-.is-style-solid span:not(.block-editor-rich-text__editable,.wp-block-post-terms__separator),
-.is-style-solid span:not(.block-editor-rich-text__editable) span
+.is-style-pill a,
+.is-style-pill span:not([class], [data-rich-text-placeholder])
 {
 	display:inline-block;
 	background-color: #f2f2f2;
@@ -69,6 +68,6 @@ a {
 	margin: 0 10px 10px 0;
 }
 
-.is-style-solid a:hover {
+.is-style-pill a:hover {
 	background:#eee;
 }
--- a/style.css
+++ b/style.css
@@ -59,8 +59,9 @@ a {
  */
 
 .is-style-solid a,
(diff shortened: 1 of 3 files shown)


carolinan:
All content needs to be inside a landmark.
The easiest way to solve this is to remove the current `<main>` and wrap everything that is between the header and footer in a new group.

style.css
```diff
@@ -52,3 +52,20 @@ a {
 .is-style-arrow-icon-details[open] > summary {
 	list-style-type: '\2192\00a0\00a0\00a0';
 }
+
+/*
+ * Styles for the taxonomy/tags on single posts
+ * https://github.com/WordPress/twentytwentyfour/issues/XX
+ */
+
+.wp-block-post-terms.taxonomy-post_tag a {
```
MaggieCabrera: Let's make this a block style so users can turn it off

MaggieCabrera (changes requested):
With the block style and the landmark issues resolved, this should be good to go

style.css
jordesign: This is now using a block style 👍

jordesign:
@MaggieCabrera Post Terms are now using a block style, and I _think_ I've correctly included the main content area in a landmark (but TBH it's the first time I've tried that - so I'm open to feedback).

I think those changes might have interfered with the review you had of some of the spacing in the template so it might need a quick re-review I'm afraid.

style.css
```diff
@@ -52,3 +52,22 @@ a {
...
 	list-style-type: '\2192\00a0\00a0\00a0';
 }
+
+/*
+ * Styles variation for post terms
+ * https://github.com/WordPress/gutenberg/issues/24956
+ */
+
+.is-style-pill a,
+.is-style-pill span:not([class], [data-rich-text-placeholder])
```
MaggieCabrera, line 62: The prefix and suffix don't need the pill style, that's for when you want to add some text before or after the tag list, so I removed the styling for those

MaggieCabrera (approved):
looks good, thank you!

style.css
jordesign: Ah awesome - thank you. I'd been trying to figure out how to not have it apply to the placeholder, but couldn't figure it out.

Also - Pill is a much better style name 😂

jordesign:
Thanks @MaggieCabrera  - with that styling confirmed for the single post - should it also be rolled across to the 'Single with sidebar' and 'Write single' templates?

MaggieCabrera:
> Thanks @MaggieCabrera - with that styling confirmed for the single post - should it also be rolled across to the 'Single with sidebar' and 'Write single' templates?

yeah, we can do that but I think Single writer needs to be removed, it's almost exactly like single

jordesign:
Agreed - it is essentially the same except the Figma design seems to not include the featured image. I'll add an issue/PR to remove it.

jordesign:
PR up in #349

jordesign:
[Issue](https://github.com/WordPress/twentytwentyfour/issues/395) and [PR](https://github.com/WordPress/twentytwentyfour/pull/396)

Issue #100, opened 2023-08-27
Template selection


MaggieCabrera:
The idea for this theme is to have three possible use cases, hence why we have multiple options for templates like home, page or single. Right now there is no easy way in WordPress to swap an existing template with an alternative provided by the theme in a way that is not manually emptying/deleting the existing template and redoing it. There are some designs explorations for this flow on https://github.com/WordPress/gutenberg/issues/36951#issuecomment-1674411656 and I've discussed this with @richtabor during WCUS and we thought that we could make use of the tools that already exist for Block Patterns.

The idea would be to register these alternate versions of the templates as patterns and make use of [templateTypes](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-patterns/#register_block_pattern) to assign the template that should use this pattern. This already works well for creation of new pages, but we are missing the flow for replacing.

richtabor:
We’d remove these templates as “templates” as well, right? I don’t think we need them front and center here, and as template types. 

We may want them to show up when you add a new page too (without header and footer).

dianeco:
Can we remove the home.html template? It conflicts with the settings saved in WordPress dashboard > Settings > Reading and can confuse the users.

richtabor:
> Can we remove the home.html template? 

Perhaps we need a quick play-by-play on the options we can use for the main page, and how those will interplay with the templateTypes selection above.

kafleg:
> Can we remove the home.html template? It conflicts with the settings saved in WordPress dashboard > Settings > Reading and can confuse the users.

Or we can change home.html to front-page.html?

maneshtimilsina:
Both `home.html` and `front-page.html` will have conflicts with the settings saved in WordPress dashboard >> Settings >> Reading as mentioned by @dianeco

richtabor:
This issue would let us swap out entire templates: https://github.com/WordPress/gutenberg/issues/51347

carolinan:
I do not understand in what way the home template conflicts with the reading settings, as mentioned above. It is the template used to display the blog.

maneshtimilsina:
@carolinan Go to Settings >> Reading and select A static page option for Your homepage displays. Assign pages for Homepage and Posts page. Check both the homepage and blog post page. I think one of them will not work as expected.

carolinan:
I made the test, and both the blog and the actual front page display the correct content.

dianeco:
[...]

PR #322, opened 2023-09-11
Add portfolio template type

richtabor:
**Description**

A step towards proving out how we can achieve alternate templates for https://github.com/WordPress/twentytwentyfour/issues/100. The flow I used in the screen recording is for new templates, but this setup will be leveraged for template replacing as well (pick the portfolio view for your existing front-page template for example). Probably also index template (but we need the replace flow to experience that). 

> We may need to have a portfolio page pattern with header and footer parts, so it's a proper full replacement — whereas currently the portfolio page pattern is just the content (like other page patterns). We may need both (which shouldn't be a heavy lift, as we're building with pattern blocks anyhow).

**Screenshots**

https://github.com/WordPress/twentytwentyfour/assets/1813435/1f0bae08-1c7c-49dc-a5ca-8946d0b25662

**Testing Instructions**

1. Activate the theme. 
2. Remove a custom front-page template if you have one. 
3. Add a new front-page template. 
4. See the portfolio home template as a choice for the template.


patterns/template-portfolio-home.php
```diff
@@ -0,0 +1,16 @@
+<?php
+/**
+ * Title: Portfolio Template
+ * Slug: twentytwentyfour/template-portfolio-home
```
MaggieCabrera: This might be bad English but should it be template-home-portfolio instead? leaving the "portfolio" bit to last as a descriptor of what kind of home template we are referring to

patterns/template-portfolio-home.php
```diff
@@ -0,0 +1,16 @@
...
+ * Template Types: front-page, index
```
MaggieCabrera: no home?

patterns/template-portfolio-home.php
```diff
@@ -0,0 +1,16 @@
...
+ * Viewport width: 1400
+ * Inserter: no
```
MaggieCabrera: I like the flexibility of building with patterns that this little thing brings, lovely

patterns/template-portfolio-home.php
richtabor: Probably? I think we should audit what we should add it for to be sure, but I can add.

patterns/template-portfolio-home.php
richtabor: Yea I like that. Will change.

MaggieCabrera (approved):

PR #323, opened 2023-09-11
Add RSVP pattern

### LINKED ISSUE #122: Block Patterns - RSVP

richtabor:
**Description**

Alternate approach to https://github.com/WordPress/twentytwentyfour/pull/317, using a similar layout implemented in the about pattern. 

Closes https://github.com/WordPress/twentytwentyfour/issues/122.

**Screenshots**

https://github.com/WordPress/twentytwentyfour/assets/1813435/e184f269-533b-4cbc-8eee-605895340181

**Testing Instructions**

1. Search for `rsvp` in the pattern inserter. 
2. Add it to the page. 
3. Check it on mobile. 

**Contributors**

@carolinan


carolinan:
🤔 The editor and the front does not match.
On the front, the large text does not align with the top of the image (it aligns in the editor).
In the editor, the large text is much further to the left, so it does not align with the smaller text and the button.

I much prefer the Figma design, where the top of the text aligns with the top of the image, and the bottom of the button aligns with the bottom of the image.

richtabor:
> I much prefer the Figma design, where the top of the text aligns with the top of the image, and the bottom of the button aligns with the bottom of the image.

Yea, but that's not a realistic expectation I would put on folks. That limits the application/use of the pattern to media that perfectly fits, rather than media that works. 

For example, this is another look you can get, just by changing the background color and setting the image to a square aspect ratio: 

CleanShot 2023-09-12 at 08 00 51

richtabor:
> 🤔 The editor and the front does not match.

Here's what I have; 1:1 in essence: 

CleanShot 2023-09-12 at 07 59 08

carolinan:
I am not able to reproduce the misalignment I saw earlier 🎉  (only tested on macOS, chrome)

carolinan:
If you align the large text to the right, the R will be at the bottom and the text read from bottom up like in Figma

richtabor:
> If you align the large text to the right, the R will be at the bottom and the text read from bottom up like in Figma

I was wondering how you did that—clever!

Done.

luminuu (approved):
LGTM!

PR #325, opened 2023-09-12
Fix PHP issue in hidden-sidebar-writer.php

carolinan:
**Description**

This is the same fix as in https://github.com/WordPress/twentytwentyfour/pull/313 but for hidden-sidebar-writer.php.

**Testing Instructions**
Create a new page.
Assign it the "home writer" template.
Save and view the front.
Confirm that there is no fatal error and the content loads.


kafleg (approved):
LGTM

juanfra (approved):

PR #331, opened 2023-09-12
Update comments area

carolinan:
**Description**

Updates the comments area to match the Figma design better.
-Moves the edit link
-Removes default underline on links (shows them on :hover)
-Adds a left border to the nested comments. **We need to consider how we want to solve this for RTL.**

Figma: https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?node-id=507%3A1539&mode=dev

**Screenshots**
Comments area

And with pagination visible:
comment area pagination

**Testing Instructions**
View a single post that has comments.

### DIFF
--- a/patterns/hidden-comments.php
+++ b/patterns/hidden-comments.php
@@ -10,32 +10,22 @@
 <!-- wp:comments {"className":"wp-block-comments-query-loop"} -->
 <div class="wp-block-comments wp-block-comments-query-loop">
 	<!-- wp:comments-title {"level":3} /-->
-
 	<!-- wp:comment-template -->
-		<!-- wp:group {"style":{"spacing":{"margin":{"top":"0","bottom":"var:preset|spacing|50"}}}} -->
-		<div class="wp-block-group" style="margin-top:0;margin-bottom:var(--wp--preset--spacing--50)">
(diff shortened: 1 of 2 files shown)


carolinan:
The edit and reply link are too close. It may be better to place them inside a row with block spacing, than using the CSS in theme.json.

MaggieCabrera:
I'm happy to skip the border altogether. Using a row block sounds good to me

carolinan:
Without border:
Nested comments without borders

carolinan:
For those who did not follow along with "Blue Note" development;
CSS generated by theme.json is **not** converted for RTL. So a left border would remain on the left side also on RTL, that is why it is problematic and why the easiest solution is to remove it.

MaggieCabrera:
> For those who did not follow along with "Blue Note" development; CSS generated by theme.json is **not** converted for RTL. So a left border would remain on the left side also on RTL, that is why it is problematic and why the easiest solution is to remove it.

I wonder if there should be an alternative CSS for rtl built in from theme.json, is there an open issue in GB?

carolinan:
Yes https://github.com/WordPress/gutenberg/issues/52353

luminuu (approved):
LGTM!

PR #334, opened 2023-09-12
English strings

### LINKED ISSUE #333: Strings that need to be translated to English
**Description**

**Patterns**

- Footer portfolio
- Centered Statement
- Heading on left (hidden, shows in Home portfolio template and pattern)

**Templates**

- Writer home

MaggieCabrera:
**Description**

Will close https://github.com/WordPress/twentytwentyfour/issues/333

### DIFF
--- a/patterns/centered-statement.php
+++ b/patterns/centered-statement.php
@@ -11,12 +11,11 @@
 <div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--50);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50)"><!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60","left":"var:preset|spacing|50","right":"var:preset|spacing|50"}},"border":{"radius":"16px"}},"backgroundColor":"base-2","layout":{"type":"default"}} -->
 <div class="wp-block-group alignwide has-base-2-background-color has-background" style="border-radius:16px;padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--50)"><!-- wp:paragraph {"align":"center","style":{"typography":{"lineHeight":"1.2","fontStyle":"normal","fontWeight":"400"}},"fontSize":"x-large","fontFamily":"cardo"} -->
 <p class="has-text-align-center has-cardo-font-family has-x-large-font-size" style="font-style:normal;font-weight:400;line-height:1.2">
-<?php
+	<?php
 		echo sprintf(
-			'%1$s<em>%2$s</em>%3$s',
-			esc_html__( 'Queste fotografie non solo hanno catturato la bellezza fisica del ', 'twentytwentyfour' ),
-			esc_html__( 'Piva Arts &amp; Crafts Hotel', 'twentytwentyfour' ),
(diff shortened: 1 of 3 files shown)


luminuu:
Quick thought: Translations that are split up so that some parts can have additional HTML tags are a PITA for translators. While it does work in some languages, it may lead to issues in other languages as the grammar needs the words to switch order in some cases. I'd suggest not to use strings like this.

MaggieCabrera:
@nilovelez care to chime in on this? Is there any way we can make it simpler for translators?

patterns/centered-statement.php
```diff
@@ -11,12 +11,11 @@
...
 <p class="has-text-align-center has-cardo-font-family has-x-large-font-size" style="font-style:normal;font-weight:400;line-height:1.2">
-<?php
+	<?php
 		echo sprintf(
-			'%1$s<em>%2$s</em>%3$s',
-			esc_html__( 'Queste fotografie non solo hanno catturato la bellezza fisica del ', 'twentytwentyfour' ),
-			esc_html__( 'Piva Arts &amp; Crafts Hotel', 'twentytwentyfour' ),
-			esc_html__( ", ma hanno anche trasmesso le emozioni e le sensazioni che l'hotel evoca nei suoi ospiti", 'twentytwentyfour' )
+			'<em>%1$s</em>%2$s',
+			esc_html__( 'Études ', 'twentytwentyfour' ),
```
sabernhardt, line 17: HTML elements can be inside translatable strings to maintain proper context. An organization/company name of "Études" without context could be extra confusing because the word is French and adopted into English. The `em` element could be inappropriate in some language(s) too.

Twenty Twenty-Two uses `wp_kses_post()` for strings like this.

```
<?php echo wp_kses_post( __( '<em>Études</em> is not confined to the past—we are passionate about the cutting-edge designs shaping our world today.', 'twentytwentyfour' ) ); ?>
```

patterns/hidden-hidden-intro-text-left.php
```diff
@@ -19,7 +19,16 @@
 <div class="wp-block-column" style="flex-basis:70%">
 
 <!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.2"}},"fontSize":"xx-large","fontFamily":"cardo"} -->
-<p class="has-cardo-font-family has-xx-large-font-size" style="line-height:1.2">Sono <em>Leia Acosta</em>, una fotografa appassionata che trova ispirazione nel catturare la bellezza fugace della vita.</p>
+<p class="has-cardo-font-family has-xx-large-font-size" style="line-height:1.2">
+	<?php
+		echo sprintf(
```
sabernhardt, line 24:
```
<?php echo wp_kses_post( __( 'I’m <em>Leia Acosta</em>, a passionate photographer who finds inspiration in capturing the fleeting beauty of life.', 'twentytwentyfour' ) ); ?>
```

nilovelez:
Totally agree with @sabernhardt
String can have simple HTML inside to keep string cohesion and make it easite to translate, specially in languages where we have a gramatical structure different than English.

Just try to avoid innecesary attributes in the HTML tags.

For example this is ok:
```php
<p class="my-cta-paragraph"><?php esc_html_e( 'Lorem ipsum <em>Hello</em> dolor sit amet.', 'texdomain' ); ?></p>
``` 

But this is unnecesary:
```php
<?php esc_html_e( '<p class="my-cta-paragraph">Lorem ipsum <em>Hello</em> dolor sit amet.</p>', 'texdomain' ); ?>
``` 

And you should try to avoid this:
```php
<p class="my-cta-paragraph"><?php esc_html_e( 'Lorem ipsum <em class="a-lot-of html-tags" id="this-is-also-an-id">Hello</em> dolor sit amet.', 'texdomain' ); ?></p>
```

MaggieCabrera:
Awesome feedback there @nilovelez @sabernhardt , let's add all this to the readme so it doesn't get lost in time and feature default theme devs apply the same rules

PR #337, opened 2023-09-12
404 pattern update

carolinan:
**Description**

- Removes colors and font sizes, to fallback on the defaults.
- Removes alignment, to fallback on the default.
- Adds a search form label (the label is still visually hidden).
- Adjusts the font size and font weight of the search block to match Figma (matches both search results page and 404).

**Testing Instructions**
View the 404 page in the editor and front.

### DIFF
--- a/patterns/hidden-404.php
+++ b/patterns/hidden-404.php
@@ -7,11 +7,10 @@
 
 ?>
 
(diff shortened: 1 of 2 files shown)


luminuu (approved):
LGTM!

PR #345, opened 2023-09-13
Updated footer for better a11y

### LINKED ISSUE #12: Add links to footer menu/items?

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/12

This PR adds headings to the footer instead of paragraphs and makes the links into nav blocks with custom placeholder content.

I think this is the best compromise between a11y concerns and the design given. We don't have a native solution for this kind of menu design yet, and I don't want to add the extra CSS for this that will be hard for a user to make it work if they want to make changes to it.

I'm not too sure about the headings here, but it was [suggested](https://github.com/WordPress/twentytwentyfour/issues/12#issuecomment-1716053412) they should be. Maybe they need to be h2s in case the page doesn't contain any h2s themselves, since the footer will be used on all pages. I'm not sure how semantic that is, happy to hear any thoughts in that regard.

/cc @mrwweb

### DIFF
--- a/patterns/footer.php
+++ b/patterns/footer.php
@@ -26,63 +26,63 @@
 <!-- wp:column {"width":"50%","style":{"spacing":{"padding":{"top":"var:preset|spacing|30"}}}} -->
 <div class="wp-block-column" style="padding-top:var(--wp--preset--spacing--30);flex-basis:50%"><!-- wp:group {"layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
 <div class="wp-block-group"><!-- wp:group {"layout":{"type":"flex","orientation":"vertical","justifyContent":"stretch"}} -->
-<div class="wp-block-group"><!-- wp:paragraph {"style":{"typography":{"fontStyle":"normal","fontWeight":"600"}},"fontSize":"medium"} -->
-<p class="has-medium-font-size" style="font-style:normal;font-weight:600">About</p>
-<!-- /wp:paragraph -->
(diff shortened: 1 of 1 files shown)


carolinan:
We need to differentiate the three `<nav>` elements.
When we use existing _menus_ in the nav block, the menu name is used for the aria-label, but in this case, the aria-label is empty.

We may need to fix this in Gutenberg.
We could try adding the ariaLabel block support to block.json, but we might run into problems with the existing aria-label.

carolinan:
If we could set submenus to always show, we could have had one nav block with 3 parent menu items and submenus...

MaggieCabrera:
> We need to differentiate the three `<nav>` elements. When we use existing _menus_ in the nav block, the menu name is used for the aria-label, but in this case, the aria-label is empty.
> 

That really sounds like a GB issue, having more than one nav block in a site is not specific to TT4, so we definitely should find a global solution

carolinan:
It works when you use an actual menu. In this case there is no reference to a menu, so there is also no menu name input field under Advanced.

(I tried adding a label to the block comment but it did not work)

carolinan:
We can merge this meanwhile.

The headings above the `<nav>` also helps, just not if you are navigating the page via landmarks or tabbing through focusable elements.

carolinan (approved):

scruffian (approved):
This looks good. I think we could simplify the stacks and convert them into simple group blocks, but that doesn't need to happen.

carolinan:
I have a PR for the fix in Gutenberg. Will add it to the list of Gutenberg issues.

PR #292, opened 2023-09-08
Left Aligned Call to Action with Image pattern adjustments

juanfra:
**Description**

Patterns tweaks from Beatriz. 
Removed padding left from first column to match designs.

Fixes https://github.com/WordPress/twentytwentyfour/issues/247

**Screenshots**

https://github.com/WordPress/twentytwentyfour/assets/252415/8ad5c23e-867f-4d9c-87d3-6d46ba17843f

PR #348, opened 2023-09-13
Reduce left padding of lists

luminuu:
**Description**

To match the Figma design for lists, they should align with the heading. This change adds a small padding, so the list items align nicely with the heading.

**Screenshots**

CleanShot 2023-09-13 at 12 20 10@2x

**Testing Instructions**

1. Create/open a post/page
2. Add an unordered or ordered list
3. Check the left alignment

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -503,7 +503,7 @@
 			"core/list": {
 				"spacing": {
 					"padding": {
-						"left": "var(--wp--preset--spacing--70)"
+						"left": "var(--wp--preset--spacing--10)"
 					}
 				}
 			},


theme.json
```diff
@@ -503,7 +503,7 @@
 			"core/list": {
 				"spacing": {
 					"padding": {
-						"left": "var(--wp--preset--spacing--70)"
```
MaggieCabrera, line 506: I wonder why this was here in the first place, are we affecting any patterns with this?

luminuu: I checked on the git blame and it was included in the initial commit. I'll check the patterns now.

luminuu: The only pattern and template with a list is the left aligned CTA with image one. Reverting back to the current state, it looks like this:
CleanShot 2023-09-13 at 13 02 52@2x

Adding in this PR makes it look like this:
CleanShot 2023-09-13 at 13 03 27@2x

There's an open PR for this pattern that tries to fix this by removing the list and adding paragraphs, that also include the checkmarks: https://github.com/WordPress/twentytwentyfour/pull/292

But tbh I'd prefer the regular dot list items here to have a semantically correct list.

MaggieCabrera: yeah, the checkmarks are in other places and we should change them to be a block style if anything, but that's a different matter, let's go with this change

MaggieCabrera (approved):

Issue #338, opened 2023-09-12
Portfolio footer: make it wide instead of full width


MaggieCabrera:
<img width="1770" alt="Screenshot 2023-09-12 at 17 40 24" src="https://github.com/WordPress/twentytwentyfour/assets/3593343/181bf3d6-97d6-4e54-a1e9-c8d29bf4a066">

Let's make this footer pattern be wide width instead of full width

jordesign:
PR up in https://github.com/WordPress/twentytwentyfour/pull/349

jordesign:
This can be closed with #349  being merged now cc/ @MaggieCabrera

PR #349, opened 2023-09-13
Updates to Portfolio Footer

jordesign:
**Description**

Addresses #338:

* Make pattern wide (by wrapping in extra Group)
* Fix invalid content warning on icon at top
* Use 'wide' style for the separator

**Screenshots**

Screenshot 2023-09-13 at 8 28 21 pm

### DIFF
--- a/patterns/footer-portfolio.php
+++ b/patterns/footer-portfolio.php
@@ -12,8 +12,8 @@
 <figure class="wp-block-image size-full is-resized"><img src="<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/icon-message.webp" alt="" style="width:40px;height:auto"/></figure>
 <!-- /wp:image -->
 
-<!-- wp:separator -->
-<hr class="wp-block-separator has-alpha-channel-opacity"/>
+<!-- wp:separator {"className":"is-style-wide"} -->
+<hr class="wp-block-separator has-alpha-channel-opacity is-style-wide"/>
 <!-- /wp:separator -->
 
 <!-- wp:columns {"style":{"spacing":{"padding":{"top":"var:preset|spacing|10"}}}} -->
@@ -74,5 +74,6 @@
 		?>
 </p>
 <!-- /wp:paragraph --></div>
+<!-- /wp:group --></div>
 <!-- /wp:group --></div>
 <!-- /wp:group -->
--- a/patterns/footer-portfolio.php
+++ b/patterns/footer-portfolio.php
@@ -7,9 +7,10 @@
  */
 ?>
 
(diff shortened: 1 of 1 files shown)


patterns/footer-portfolio.php
```diff
@@ -7,13 +7,14 @@
  */
 ?>
 
-<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"7vh","bottom":"7vh"}}}} -->
-<div class="wp-block-group alignwide" style="padding-top:7vh;padding-bottom:7vh"><!-- wp:image {"width":"40px","height":"auto","sizeSlug":"full","linkDestination":"none"} -->
-<figure class="wp-block-image size-full is-resized"><img src="<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/icon-message.webp" alt="" style="width:40px;height:auto"/></figure>
```
MaggieCabrera, line 12: the invalid warning with the image was a change that is not on 6.3 but will be on 6.4

MaggieCabrera: basically aspect ratio

patterns/footer-portfolio.php
```diff
@@ -7,13 +7,14 @@
...
+<!-- wp:group {"align":"wide","layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignwide"><!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"7vh","bottom":"7vh"}}}} -->
+<div class="wp-block-group alignwide" style="padding-top:7vh;padding-bottom:7vh"><!-- wp:image {"width":40,"sizeSlug":"full","linkDestination":"none","className":"is-resized"} -->
```
MaggieCabrera: is there a way to do this without two group blocks? how do the other footers handle this?

patterns/footer-portfolio.php
jordesign: Ah of course - I've reverted to the previous version for that figure.

patterns/footer-portfolio.php
jordesign: Hmmmm so...

- Writer Footer is _also_ full width - but because it is center aligned, you can't tell.
- main Footer has the Columns block for layout - which has the 'wide' width applied.

Looks like I _might_ be ble to do the same kind of thing with the inner elements of this footer, with the exception of the icon/image as I can't set a width on that block.

jordesign:
Updated PR has wide width for all elements without additional containing blocks _except_ image at top. Very open to suggestions of how to achieve that without an additional group.

MaggieCabrera (approved):
You were right, we needed the extra wrapper, let's bring this one in

MaggieCabrera (approved):

PR #350, opened 2023-09-13
Pullquote: make cite sans-serif

luminuu:
**Description**

This PR adds some code to make the pullquote citation sans-serif as designed in the Figma file.

**Screenshots**

CleanShot 2023-09-13 at 12 47 02@2x

**Testing Instructions**

1. Create/open a post/page
2. Add pullquote
3. Add citation

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -630,9 +630,6 @@
 						"width": "0px"
 					}
 				},
-				"color": {
-					"background": "var(--wp--preset--color--tertiary)"
-				},
 				"elements": {
 					"cite": {
 						"typography": {
--- a/theme.json
+++ b/theme.json
@@ -633,6 +633,15 @@
 				"color": {
 					"background": "var(--wp--preset--color--tertiary)"
 				},
+				"elements": {
+					"cite": {
+						"typography": {
+							"fontFamily": "var(--wp--preset--font-family--system-font)",
+							"fontSize": "var(--wp--preset--font-size--medium)",
+							"fontStyle": "normal"
+						}
+					}
+				},
 				"spacing": {
 					"padding": {
 						"bottom": "var(--wp--preset--spacing--50)",


carolinan:
I will approve this, but should the default pull quote also have a white background color? I think the background color is wrong, but it can be fixed separately.

https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?node-id=492%3A1684&mode=dev

carolinan (approved):

MaggieCabrera:
> I will approve this, but should the default pull quote also have a white background color? I think the background color is wrong, but it can be fixed separately.
> 
> https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?node-id=492%3A1684&mode=dev

looks like it does, and a border radius

luminuu:
I was actually looking at the pullquotes here in that long example post: https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=565-9983&mode=design&t=hoFtF7QzImxpBnlX-0

These do not have a background color and border radius. Maybe @beafialho can clarify which design we should use as default for the pullquotes?

beafialho:
Thanks for checking, I just tested the PR and the default quote, the plain one and pullquote styles are correct 👍🏼

Would it be possible to add a dash to quote citation, as in the Figma design?

| Currently      | Figma |
| ----------- | ----------- |
|   Captura de ecrã 2023-09-13, às 16 02 07  | Captura de ecrã 2023-09-13, às 16 02 21  |

carolinan:
This color doesn't exist in the theme though? So theme.json still needs to be updated. It doesn't have to be in this PR. 

```
"color": {
"background": "var(--wp--preset--color--tertiary)"
				},
```

MaggieCabrera:
> This color doesn't exist in the theme though? So theme.json still needs to be updated. It doesn't have to be in this PR.
> 
> ```
> "color": {
> "background": "var(--wp--preset--color--tertiary)"
> 				},
> ```

looks like we missed it when we changed the color names?

MaggieCabrera (approved):

PR #352, opened 2023-09-13
Add border to default button to match size with outlined button

luminuu:
**Description**

This PR adds a border for the default buttons, which currently do not have one. Putting a default button and an outlined button side by side, there's a minor difference in size, as the outlined button has the border, the default has not.

**Screenshots**

Before:
CleanShot 2023-09-13 at 13 26 07@2x

After:
CleanShot 2023-09-13 at 13 25 33@2x

**Testing Instructions**

1. Create/open a post/page
2. Add a pattern with both buttons or create a new buttons block, add two buttons with one being default and the other outlined

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -780,7 +780,24 @@
 					}
 				},
 				"border": {
-					"radius": ".33rem"
+					"radius": ".33rem",
+					"color": "var(--wp--preset--color--contrast)",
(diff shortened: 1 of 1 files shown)


carolinan (approved):
The difference without the extra border is apparent, this PR solves it, thank you.

PR #357, opened 2023-09-13
Add Theme Style Variation: Onyx

### LINKED ISSUE #10: Style Variations

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/10

This PR creates the 3 style variations. We have a problem with the way we've built patterns, and some of them apply a specific font family to the markup, making the block styles inconsistent. I don't know if we can fix all of them (in some cases we are applying the Cardo font family to paragraphs so they look like they are headings when they are not). 

Screenshot 2023-09-13 at 18 42 24

My first instinct is to create a custom variable called heading font and another called base font or something like that, to use them in patterns and then map them in theme.json. This would be the perfect opportunity for refs to work on block markup.

EDIT: This PR paired with https://github.com/WordPress/twentytwentyfour/pull/366 should cover the patterns too


MaggieCabrera:
@richtabor @scruffian what do you think?

MaggieCabrera:
Related: https://github.com/WordPress/gutenberg/issues/49279#issuecomment-1480201164

scruffian:
We could also use things like:
https://github.com/WordPress/gutenberg/issues/43347
https://github.com/WordPress/gutenberg/issues/27233
https://github.com/WordPress/gutenberg/issues/40318

richtabor:
Maybe "body" and "heading" font works better; so variations can register different fonts and they're hooked up as expected.

MaggieCabrera:
I think a variation of paragraph could work as well, are we overdoing it with all the block styles?

richtabor:
I don't think we're overdoing it, but I'm a bit hesitant adding a style to such a prominent block. We can try it in a quick PR, but I'm not sure it's necessary.

MaggieCabrera:
@richtabor I gave it a try [here](https://github.com/WordPress/twentytwentyfour/pull/366) and I was sadly surprised that you can only change styles via theme.json to default variations and not theme created ones.

richtabor:
This should do the trick: https://github.com/WordPress/twentytwentyfour/pull/367

richtabor:
Also, we need to update accent colors to have a darker luminosity. I can do that. 

CleanShot 2023-09-14 at 13 44 53
CleanShot 2023-09-14 at 13 44 49
CleanShot 2023-09-14 at 13 44 41

richtabor:
Let's focus on the first variation before doing all the others.

richtabor:
I made some adjustments, focusing on color application. Notably: 

1. Using consistent "slugs" on duotone and gradients ensures that the application of those attributes persist with different variations. That is, duotone-2 is styled by both variations — and still works when used throughout content. Also, the labels are still applicable to the actual color used, i.e. "Dark gray and cinnamon" 

Notice how the duotone works for both: 

https://github.com/WordPress/twentytwentyfour/assets/1813435/42d43250-f0e8-4d60-ad19-e6ca7318432f


2. Removed the additional style variations to focus on just the one. Let's get this solid, then replicate.

richtabor (approved):
LGTM as an initial push. We'll want to circle back with font family adjustments, but let's merge this before potential conflicts.

PR #365, opened 2023-09-14
Try: add a block style for separator blocks that turns them into asterisks

MaggieCabrera:
**Description**

Alternative to https://github.com/WordPress/twentytwentyfour/pull/51

Uses SVG and clip path to turn the separator into an asterisk. This is a proof of concept, we got to figure out:

- how to tinker with the alignment of the separator
- If we are happy with using the hr for this particular case
- if using px units to lock in the size of the icon is ok

**Screenshots**

This screenshot shows: 
- the old paragraph block with the glyph
- 3 separator blocks with different colors applied to it

Screenshot 2023-09-14 at 10 58 59

**Testing Instructions**

Add a separator block and check the asterisk block style for it

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -80,6 +80,13 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 				'label' => __( 'Pill', 'twentytwentyfour' ),
 			)
 		);
+		register_block_style(
+			'core/separator',
+			array(
+				'name'  => 'asterisk',
+				'label' => __( 'Asterisk', 'twentytwentyfour' ),
+			)
+		);
 	}
 endif;
 
--- a/patterns/feature-grid.php
+++ b/patterns/feature-grid.php
@@ -9,9 +9,11 @@
 
 <!-- wp:group {"align":"full","style":{"spacing":{"margin":{"top":"0","bottom":"0"},"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50","left":"var:preset|spacing|50","right":"var:preset|spacing|50"}}},"backgroundColor":"base-2","layout":{"type":"constrained"}} -->
 <div class="wp-block-group alignfull has-base-2-background-color has-background" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--50);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50)"><!-- wp:group {"style":{"spacing":{"blockGap":"0px"}},"layout":{"type":"flex","orientation":"vertical","justifyContent":"center"}} -->
-<div class="wp-block-group"><!-- wp:paragraph {"align":"center","textColor":"base-3","fontSize":"large"} -->
-<p class="has-text-align-center has-base-3-color has-text-color has-large-font-size">✳︎</p>
-<!-- /wp:paragraph -->
+<div class="wp-block-group">
+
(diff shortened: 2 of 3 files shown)


carolinan:
This is working pretty well for me. Some notes:

- The background color with the clip is working 🎉 
- VoiceOver with Chrome is announcing "Horisontal splitter" instead of the asterisk.
- It was unexpected for me that setting full width removed the symbol while the Asterisk block style was selected.
If full width was allowed it could be used to position the symbol to the left. Without it, positioning is more difficult.

I think this can be solved by removing the `:not` from the stylesheet, and adding the class name of the new style variation to the custom CSS field in theme.json. I mean, including it here:

```
			"core/separator": {
				......
				"css": " &:not(.is-style-wide):not(.is-style-dots):not(.alignwide):not(.alignfull){width: var(--wp--preset--spacing--60)}"
			},
```

About the pixel value, I think this only matters on very large screens like if someone would use it on a larger TV, billboard type setting? Then it might look too small?

MaggieCabrera:
I learned that:

- We can actually use SVG on the image block via patterns, the editor only checks the extension if we are inserting
- We can't really control the color of the SVG if inserted that way, which breaks the style variations...
- I worked on this with @jeryj and I'm not sure an hr is less annoying than the asterisk paragraph for screen readers???

I'm not sure where to go with this.

carolinan:
Maybe use the HR, but reduce the number of times it is used in the pattern? Perhaps the grid has 3 sections not 6.

richtabor:
@joedolson What do you think of this angle to resolve https://github.com/WordPress/twentytwentyfour/issues/41?

joedolson:
An `hr` element is announced by screen readers, so I think that's only viable if the asterisk is a legitimate representation of a section break. Most screen readers announce it as 'separator'; VoiceOver announces it as 'horizontal splitter' or 'horizontal separator' depending on browser.  So it would still be fairly verbose. The announcement is meaningful, however, as long as the semantics match the usage.

richtabor:
It's more display than an actual separation of content:

CleanShot 2023-09-17 at 21 34 03

joedolson:
Then it's basically the same problem as before, just a different reading. So I don't think that's a workable solution

carolinan:
Could we try CSS only, custom block styles for the heading block?

MaggieCabrera:
> Could we try CSS only, custom block styles for the heading block?

we would need two. One for left aligned and one for centered. I'm not too happy about it either but it might be the only way. We made it work with an image block and SVG but the problem that had is that the color doesn't adapt to the style variation and since the SVG is not embedded in the HTML, CSS changes won't affect it. @jeryj do you have any other ideas for this?

jeryj:
I'll explore it more again today. To summarize the ideal requirements:

- Asterisk shows in the editor
- Can change color via block settings
- Can use alignment (center/left/right) via block settings
- Doesn't have any screen reader announcement, as it's a decorative element, not semantic

Anything else to add to this?

MaggieCabrera:
>     * Can use alignment (center/left/right) via block settings
> 

Doesn't need to be block settings, but I'd rather not have 2 block variations for the user to see that only have centering as a difference between them

colorful-tones:
@MaggieCabrera @jeryj I just pushed up #448, which I believe addresses all the requirements:

- allow user to select background color
- allow user to choose alignment
- uses Separator block and `is-style-asterisk` block style

jeryj:
I have the requirements met in https://github.com/WordPress/twentytwentyfour/pull/449 using a Group block wrapping a Spacer block with an Asterisk clip path in the CSS, so there are no semantics, and the text color and alignment can be used. It works quite well for allowing styling using the group block styles and settings.

PR #366, opened 2023-09-14
Try: add heading block style to paragraph

MaggieCabrera:
This PR creates a paragraph style variation so that the font family values can be altered by theme.json for the [style variations](https://github.com/WordPress/twentytwentyfour/pull/357) to actually work as intended.

Why we use the css property of theme.json for this: https://github.com/WordPress/gutenberg/issues/49550

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -84,7 +84,7 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 			'core/paragraph',
 			array(
 				'name'  => 'heading',
-				'label' => __( 'Heading', 'twentytwentyfour' )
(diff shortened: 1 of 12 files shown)


carolinan:
I think styling a paragraph as a heading -even only by the name - is a bad idea because it can give users the idea that headings and paragraphs are interchangeable.

We should not make it easier for the user to make mistakes that can cause problems both for SEO and accessibility.

I don't have a great alternative at the moment... 
I think we could use a class name without the confusing block style variation, or, use a custom css property for the alternate font family in theme.json + theme style variations.-

MaggieCabrera:
I agree, this was just an exploration, but I'm not a fan either

PR #367, opened 2023-09-14
Use font slugs

richtabor:
**Description**

Uses `heading` as a font family slug, instead of `cardo` so that there is a default heading font that can be changed by variations, and patterns can inherit the new style's `heading` font. And the same for `body` font. 

This is how colors and spacing works, so that variations can change them holistically. 

**Screenshots**
No visual change. The font family is still named "Cardo" in the UI. 

CleanShot 2023-09-14 at 13 35 52

### DIFF
--- a/patterns/about-2.php
+++ b/patterns/about-2.php
@@ -14,8 +14,8 @@
 <div class="wp-block-column is-vertically-aligned-stretch" style="flex-basis:50%"><!-- wp:group {"style":{"dimensions":{"minHeight":"100%"}},"layout":{"type":"flex","orientation":"vertical","justifyContent":"stretch","verticalAlignment":"space-between"}} -->
 <div class="wp-block-group" style="min-height:100%">
 
-<!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.2"}},"fontSize":"x-large","fontFamily":"cardo"} -->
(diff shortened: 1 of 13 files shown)


MaggieCabrera:
This works nicely but then when I try and build the sans serif variations I need to apply the same font to both body and heading like this:

```
{
       "settings": {
"typography": {
			"fontFamilies": [
				{
					"fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
					"name": "System Font",
					"slug": "body"
				},
				{
					"fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
					"name": "System Font",
					"slug": "heading"
				}
			]
		} 
        }
}
```

And that turns into this on the dropdown:

Screenshot 2023-09-15 at 09 41 52

richtabor:
Yea I don’t know there’s a way around that. We could adapt the design to use a different heading font, like Inter, to avoid.

MaggieCabrera:
> Yea I don’t know there’s a way around that. We could adapt the design to use a different heading font, like Inter, to avoid.

I think that is a nice idea

jasmussen:
Inter is lovely and close to the system font, that could work fine IMO.

MaggieCabrera (approved):

MaggieCabrera:
Let's bring this in and go with Inter for the style variation headings

scruffian (approved):
Great work!

Issue #351, opened 2023-09-13
Index templates


MaggieCabrera:
According to the figma, we have 3 different designs for Index/Archive pages. Let's follow the same structure as we did for the home pages to develop patterns for the portfolio/writer options of these templates. Part of these are also reused on the search template, let's make everything modular using patterns.

**Default Archive:**

![Archive Template](https://github.com/WordPress/twentytwentyfour/assets/3593343/a45bc854-012e-4ea0-bef3-ec304af63e13)

**Portfolio archive:**

![Portfolio Archive](https://github.com/WordPress/twentytwentyfour/assets/3593343/d198b4db-9045-48e6-a561-951e5c185711)

**Writing Archive:**

![Writing Archive](https://github.com/WordPress/twentytwentyfour/assets/3593343/bbecb93e-536e-4b90-b95d-98ff26330798)

carolinan:
We need to be careful of the naming here:  Should we use the template hierarchy or not?
For example archive-portfolio would always be used for a post type called portfolio.

MaggieCabrera:
> We need to be careful of the naming here: Should we use the template hierarchy or not? For example archive-portfolio would always be used for a post type called portfolio.

This is why we should make these into patterns and not templates. We have a flow now to replace archive.html into something else via the patterns that we have. Adding the template is confusing.

carolinan:
I don't think it is possible to do a query variation where only posts with featured images are included, is it?

PR #368, opened 2023-09-15
Add portfolio archive template with custom query variation

carolinan:
**Description**

- Adds a full page pattern for the portfolio archive template, includes header and footer template parts. 
- Registers a block variation of the query loop that only includes posts with featured images.

For https://github.com/WordPress/twentytwentyfour/issues/351
Figma: https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?node-id=492%3A1303&mode=dev

**Screenshots**
64 local__cat=17

**Testing Instructions**
Remove all the blocks in an archive template like home, archive or index, and insert the portfolio archive pattern.

### Todo:
I am not sure if I picked the correct aspect ratio for the featured images.
Query for only featured images needs fixing


carolinan:
@greenshady Hi! I am testing a custom query variation for Twenty Twenty-Four, where only posts with featured imaqes should show. I can't quiet get it to work on the front. The variation is printing empty li elements for posts that does not have featured images.

This is an experiment, I understand if it is decided that a block variation is not suitable for a bundled theme.

assets/js/block-variation.js
```diff
@@ -0,0 +1,34 @@
+wp.blocks.registerBlockVariation( 'core/query', {
+	name: 'twentytwentyfour/featured-image-query',
+	title: 'Featured Images',
+	description: 'List of posts with featured images',
```
carolinan: Are these translation ready?

assets/js/block-variation.js
```diff
@@ -0,0 +1,34 @@
...
+			'core/post-template',
+			{ layout: {
+				type: 'grid',
+				columnCount: 3
+			} },
+			[
+				[
+					'core/post-featured-image',
+					{
+						aspectRatio: 2/3,
```
carolinan, line 33: I'm not sure this is working correctly perhaps it should be `aspectRatio: '2/3'`,

carolinan:
Let me ping the correct username 😆 @justintadlock

functions.php
```diff
@@ -85,3 +85,61 @@ function twentytwentyfour_block_styles() {
...
+		get_template_directory_uri() . '/assets/js/block-variation.js',
+		array( 'wp-blocks' ),
+		wp_get_theme()->get( 'Version' ),
+		true
+	);
+}
+add_action( 'enqueue_block_assets', 'twentytwentyfour_register_block_variation' );
+
+function twentytwentyfour_rest_filter_query( $args ) {
+	// Only return posts with featured images.
```
carolinan: Need to figure out how to make sure this only affects the variation :)

assets/js/block-variation.js
MaggieCabrera: Looking at https://developer.wordpress.org/block-editor/reference-guides/block-api/block-variations/ and googling some code examples, you want this to be `__( 'List of posts with featured images', 'twentytwentyfour' )`

MaggieCabrera:
I think this is interesting to explore on a default theme, as long as we keep it clean and simple, but I'd love others opinions. /cc @richtabor 

I like the idea that default themes started showcasing how much the editor can do without adding the extra work but now we can also explore how much a little extra can bring to the theme too. We just don't wanna lean on adding functionality where it shouldn't be.

MaggieCabrera:
I understand this is still a WIP, do you think we could have a separate PR for the templates that are indeed ready to be merged? so we can have designers review them on Monday while you work on this and we discuss it

templates/archive.html
```diff
@@ -4,37 +4,10 @@
...
-		<!-- wp:query-pagination -->
-			<!-- wp:query-pagination-previous /-->
-			<!-- wp:query-pagination-numbers /-->
-			<!-- wp:query-pagination-next /-->
-		<!-- /wp:query-pagination -->
-	</div>
-	<!-- /wp:group -->
+<!-- wp:group {"tagName":"main","style":{"spacing":{"blockGap":"0","margin":{"top":"0"}}},"layout":{"type":"default"}} -->
+<main class="wp-block-group" style="margin-top:0">
+	<!-- wp:pattern {"slug":"twentytwentyfour/archive"} /-->
```
MaggieCabrera: Since the only difference between archive and index is the existence of the query title, can we have this be index.html instead? The query title shouldn't show for the index template (and the index template that we have right now in the repo is incorrect if we look at the design)

carolinan: do we delete archive? it falls back to index.

carolinan: the index still needs an H1

MaggieCabrera: yeah the idea would be to delete archive. Looking at the design, I'm not sure what the h1 should be! What do you think?

carolinan: There is no way to conditionally add an H1 to index and not to the archives if archive.html is deleted.

MaggieCabrera: yeah, I know, but what would we add as h1 anyway?

carolinan: _Watch, Read, Listen_

MaggieCabrera: that makes sense. Can we movew the query title outside the query loop and just have a template for each that does heading + pattern for the query loop, then?

carolinan: Yeah

carolinan: https://github.com/WordPress/twentytwentyfour/pull/374

functions.php
```diff
@@ -84,3 +84,62 @@ function twentytwentyfour_block_styles() {
...
+		'twentytwentyfour/featured-image-query' === $parsed_block['attrs']['namespace']
+	) {
+		add_filter(
+			'query_loop_block_query_vars',
+			function( $query, $block ) use ( $parsed_block ) {
+				// Only return posts with featured images.
+				$query['meta_query'] = array(
+					array(
+						'key'     => '_thumbnail_id',
+						'compare' => 'EXISTS',
```
carolinan, line 135: When searching through WP trunk, I found that 'compare' => 'EXISTS', is only used in the tests, is there a reason for this? Are there better options?

richtabor:
Interesting. :) 

How is this different from the alternate portfolio template and pattern?

carolinan:
@richtabor Which is that? The portfolio home? https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?node-id=536%3A4185&mode=dev

functions.php
```diff
@@ -84,3 +84,65 @@ function twentytwentyfour_block_styles() {
...
+		array( 'wp-blocks', 'wp-i18n' ),
+		wp_get_theme()->get( 'Version' ),
+		true
+	);
+}
+add_action( 'enqueue_block_assets', 'twentytwentyfour_register_block_variation' );
+
+function twentytwentyfour_rest_filter_query( $args, $request ) {
+	$has_featured_image = $request->get_param( 'hasFeaturedImage' );
+	if ( $has_featured_image ) {
```
carolinan, line 110: The custom `hasFeaturedImage` doesn't seem to exist, for example it is not included if log the content of `get_params()`.
So the condition is never true and the result includes posts without featured images.

assets/js/block-variation.js
```diff
@@ -0,0 +1,40 @@
...
+	description: __( 'List of posts with featured images' ),
+	attributes: {
+		namespace: 'twentytwentyfour/featured-image-query',
+		query: {
+			perPage: 6,
+			postType: 'post',
+			order: 'desc',
+			orderBy: 'date',
+			inherit: false,
+			hasFeaturedImage: true,
```
carolinan, line 17: @ryanwelcher Do you have time look at this query loop variation? I can't get this custom query parameter (hasFeaturedImage) to work.
Yes I read the docs, Justin's article, your plugin code, _and_ watched the hallway hangout video 😆

ryanwelcher: Sure thing! I'll have a look when I am back in the office on Monday

richtabor:
Yes. We have this page pattern already here, which uses featured images. https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/page-05-portfolio-home.php

richtabor:
We should abstract that pattern as a template, like we did here: https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/template-home-portfolio.php

carolinan:
@richtabor The portfolio home design is different than the archive and search.

richtabor:
Ah, I see. The screenshot was throwing me off; looks like our existing "portfolio" home pattern.

justintadlock:
> I am testing a custom query variation for Twenty Twenty-Four, where only posts with featured imaqes should show. I can't quiet get it to work on the front. The variation is printing empty li elements for posts that does not have featured images.

Are you still running into this problem? I'm testing against trunk and Gutenberg 16.6 and am not seeing this. It looks to be working as expected.

carolinan:
> > I am testing a custom query variation for Twenty Twenty-Four, where only posts with featured imaqes should show. I can't quiet get it to work on the front. The variation is printing empty li elements for posts that does not have featured images.
> 
> Are you still running into this problem? I'm testing against trunk and Gutenberg 16.6 and am not seeing this. It looks to be working as expected.

Yes I am seeing posts without featured images printed like:
```
<li class="wp-block-post post-1241 post type-post status-publish format-standard hentry category-classic category-uncategorized tag-sticky-2 tag-template">
</li>
```

OH wait its because it is **sticky?**

carolinan:
Removing the sticky helped with the problem I was seeing with the variation: but not with the pattern issue.

carolinan:
After discussions and troubleshooting with help from Ryan Welcher, Fränk Klein and Ari Stathopoulos, I have not been able to make the block variation work _**well enough.**_

It works OK as a block variation when the query loop inherit setting is set to false. But since it is meant to be an archive and search result that will not work.
When it is set to true, it does not display the correct posts in the editor or front.

## The problem
When the post template only contains the featured image, it still outputs the list item for posts that do not have featured images. This leaves large gaps in the list of blog posts.

## Options

1. Ignore this and let there be gaps when the user forgot to add the image. That is the approach taken with the existing pattern with the image offset.
2. Hide the list items with a small amount of CSS. **Side effect:** There may be archive pages that doesn't output any content because all posts in the result are missing images.
3. Use a PHP filter for the front of the website that removes the list items. **Side effect:** There may be archive pages that doesn't output any content because all posts in the result are missing images.
4. Try to get the featured image place holder feature into 6.4

luminuu:
I assume you're referring to featured images of a post here, not images within the content of a post, correct?

I'd prefer either option 1 or 4.

carolinan:
Yes, featured images

carolinan:
Lets try to get pattern + templates in for 1. That is the simplest way right now.

( Of course there are options to show texts but it would be a totally different design. )

carolinan:
Also, there was feedback that "portfolio" is confusing and that it could be called  "archive with featured images" or similar.

MaggieCabrera:
Yeah, I wonder if there's a bug to report after you went through all this effort? It was really great that you tried this. I'd also say go for 1 but I'm not sure I understand how placeholders would be a solution? what is the issue related?

carolinan:
I tried searching for the Gutenberg issue quickly, I need some more time to find it.

basically, it is about setting an image that is used when a post type item does not have a featured image. A fallback.

carolinan:
Yes I plan to submit a bug report or a question once I am able to put words to describe the bugs 😆

MaggieCabrera:
> I tried searching for the Gutenberg issue quickly, I need some more time to find it.
> 
> basically, it is about setting an image that is used when a post type item does not have a featured image. A fallback.

Oh, I thought it was something just for the editor. That sounds like a great idea.

carolinan:
https://github.com/WordPress/gutenberg/issues/34865

luminuu:
Linked in that issue there's a workaround via the functions.php: https://github.com/WordPress/gutenberg/issues/32447#issuecomment-854949102

carolinan:
> Linked in that issue there's a workaround via the functions.php: [WordPress/gutenberg#32447 (comment)](https://github.com/WordPress/gutenberg/issues/32447#issuecomment-854949102)

Interesting

carolinan:
We are going to come back to the following aren't we:
1. Selecting an image that doesn't look like the site is half broken (designers)
2. Making that match with all style variations

luminuu:
If it's possible to conditionally show fallback images based on the selected style variation, you'd only need one master fallback image in most likely grey colors, so it fits with every style variation. Or just pick that one in general.

carolinan:
It would be possible IF this can be merged before the RC today.. https://github.com/WordPress/gutenberg/pull/51512

MaggieCabrera:
> It would be possible IF this can be merged before the RC today.. [WordPress/gutenberg#51512](https://github.com/WordPress/gutenberg/pull/51512)

I'm not sure I understand what you need to do here. Can't you target the variation with CSS now?

luminuu:
Nope, that's currently not possible to target a style variation via CSS.

MaggieCabrera:
I'm thinking that since this is not the main archive pattern and it's something that users will choose to use because it fits their website, it's fine if we go with option 1 and let the user fix their featured images. It would be nice to have this kind of functionality be easier to include on a website. 
I'm also thinking how pagination really doesn't work with this implementation, which is something most people with this kind of page would want to have.

justintadlock:
I would avoid Query Loop variations altogether if this is meant for the main query on archive pages. This is because the main query is going to take hold here. 

As just a general pattern, I think it's OK to just let users worry about featured images if nothing can be done about the issue with variations + patterns or there's no good way to do default featured images.

> Linked in that issue there's a workaround via the functions.php: WordPress/gutenberg#32447 (comment)

I was just going to suggest this. :) 

> I'm not sure I understand what you need to do here. Can't you target the variation with CSS now?

You can't target a variation with CSS. But you can potentially build something via `settings.custom` in `theme.json`. All it does is create CSS custom properties, which can be overwritten in style variations.

carolinan:
Has anyone tried using the post_thumbnail_html filter? I assume we want to insert the markup for the featured image block with the correct src: including the figure, link, class names, style, sizes and srcset.

luminuu:
@carolinan I don't think anyone has tried this yet.

PR #369, opened 2023-09-15
Block style for checkmarks

### LINKED ISSUE #46: Home page pattern - paragraphs are used for lists

MaggieCabrera:
**Description**

Alternative to https://github.com/WordPress/twentytwentyfour/pull/150
Closes https://github.com/WordPress/twentytwentyfour/issues/46

I made this PR because rebasing the older one was a little bit of a mess

**Screenshots**

Screenshot 2023-09-15 at 10 51 15
Screenshot 2023-09-15 at 10 51 10

**Testing Instructions**

Insert a list block and toggle the checkmarks style

### DIFF
--- a/style.css
+++ b/style.css
@@ -71,3 +71,15 @@ a {
 .is-style-pill a:hover {
 	background:#eee;
 }
+
+/*
+ * Styles for the custom checkmark list block style
+ * https://github.com/WordPress/gutenberg/issues/51480
+ */
+ul.is-style-checkmark-list {
+	list-style-type: "\2713";
+}
+
+ul.is-style-checkmark-list li {
+	padding-inline-start: 1ch;
+}
--- a/functions.php
+++ b/functions.php
@@ -80,6 +80,13 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 				'label' => __( 'Pill', 'twentytwentyfour' ),
 			)
 		);
+		register_block_style(
+			'core/list',
+			array(
+				'name'         => 'checkmark-list',
+				'label'        => __( 'Checkmarks', 'twentytwentyfour' ),
+			)
+		);
 	}
 endif;
 
--- a/functions.php
+++ b/functions.php
@@ -83,8 +83,8 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 		register_block_style(
 			'core/list',
 			array(
-				'name'         => 'checkmark-list',
-				'label'        => __( 'Checkmarks', 'twentytwentyfour' ),
+				'name'  => 'checkmark-list',
+				'label' => __( 'Checkmark', 'twentytwentyfour' ),
 			)
 		);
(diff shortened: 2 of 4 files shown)


carolinan:
Should the line height be in the CSS file too instead of in the pattern markup?

MaggieCabrera:
> Should the line height be in the CSS file too instead of in the pattern markup?

No, I think it's relevant for the user to change it in the editor if they change typography and need to adjust (try writing a list item that wraps to multiple lines)

style.css
```diff
@@ -71,3 +71,15 @@ a {
...
+/*
+ * Styles for the custom checkmark list block style
+ * https://github.com/WordPress/gutenberg/issues/51480
+ */
+ul.is-style-checkmark-list {
+	list-style-type: "\2713";
+}
+
+ul.is-style-checkmark-list li {
+	padding-inline-start: 1ch;
```
richtabor, line 84: Why `ch`?

richtabor:
> No, I think it's relevant for the user to change it in the editor if they change typography and need to adjust (try writing a list item that wraps to multiple lines)

I agree. The checkmark looks fine regardless of line height.

richtabor (approved):
LGTM

PR #371, opened 2023-09-15
Pattern: Writer Archive Template

carolinan:
**Description**

Adds a full page pattern for the writer archive template.

For https://github.com/WordPress/twentytwentyfour/issues/351
Figma: https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?node-id=492%3A1303&mode=dev

**Screenshots**
64 local__cat=1 (1)

**Testing Instructions**
Delete all blocks in archive.html, and add
`<!-- wp:pattern {"slug":"twentytwentyfour/archive-writer"} /-->`
View an archive page.


carolinan:
Note that if you insert the pattern via the site editor, the header and footer will not load.

theme.json
```diff
@@ -650,6 +650,9 @@
 					"lineHeight": "1.5"
 				}
 			},
+			"core/query-title": {
+				"css": "& span {font-style: italic;}"
```
MaggieCabrera, line 654: @carolinan what do you think about this?

MaggieCabrera:
I made a few changes so it looks closer to the design

theme.json
carolinan: It is the only way to style the query object name (like category name). :)

MaggieCabrera: yeah, I just feel like it looks so much better like this

MaggieCabrera:
@carolinan if all is good with you I'd like to merge this one in

MaggieCabrera (approved):

PR #375, opened 2023-09-15
Allow the home pattern to load in the site editor

scruffian:
**Description**
Adds the wp_template type to the home pattern to allow it to be loaded in the site editor, as suggested in https://github.com/WordPress/gutenberg/pull/41842#issuecomment-1239547826.

**Screenshots**
Screenshot 2023-09-15 at 12 30 32

**Testing Instructions**
Open the site editor and check that you can see the home page template.

**Contributors**

### DIFF
--- a/patterns/page-01-business-home.php
+++ b/patterns/page-01-business-home.php
@@ -5,7 +5,7 @@
  * Categories: about
  * Keywords: page, starter
  * Block Types: core/post-content
- * Post Types: page
+ * Post Types: page, wp_template
  * Viewport width: 1400
  */
 ?>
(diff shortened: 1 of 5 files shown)


patterns/page-01-business-home.php
```diff
@@ -5,7 +5,7 @@
  * Categories: about
  * Keywords: page, starter
  * Block Types: core/post-content
- * Post Types: page
+ * Post Types: page, wp_template
```
MaggieCabrera, line 8: let's add this to page-portfolio-home too

carolinan (approved):
Smart! Works well for me.

patterns/page-01-business-home.php
richtabor: Should probably be on all the page patterns.

PR #374, opened 2023-09-15
Update default index, archive and search

carolinan:
**Description**
Adds an archive pattern for the default index, archive and search.

For https://github.com/WordPress/twentytwentyfour/issues/351
Figma: https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?node-id=492%3A1303&mode=dev

**Testing Instructions**

To test the index, first delete home.html, then view the front of the website.
View an archive to test the archive template.
View the search results page to test the search.

PR #377, opened 2023-09-15
Add full page patterns for Writer: Index, search, archive

carolinan:
**Description**
Adds a pattern for a query that is then re-used for index, archive and search for the "Writer" use case.

Figma: https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?node-id=615%3A11228&mode=dev

**Screenshots**

**Testing Instructions**
Delete home.html
Delete all blocks in index.html and add `<!-- wp:pattern {"slug":"twentytwentyfour/index-writer"} /-->`.
View the index.

Delete all blocks in archive.html and add `<!-- wp:pattern {"slug":"twentytwentyfour/archive-writer"} /-->`.
View an archive page.

Delete all blocks in search.html and add `<!-- wp:pattern {"slug":"twentytwentyfour/search-writer"} /-->`.
View a search results page.

### DIFF
--- a/patterns/archive-writer.php
+++ b/patterns/archive-writer.php
@@ -10,7 +10,7 @@
 
 <!-- wp:group {"style":{"position":{"type":"sticky","top":"0px"}},"layout":{"type":"default"}} -->
 <div class="wp-block-group">
-	<!-- wp:template-part {"slug":"header"} /-->
+	<!-- wp:template-part {"slug":"header","theme":"twentytwentyfour"} /-->
 </div>
 <!-- /wp:group -->
 
@@ -25,4 +25,4 @@
 </main>
 <!-- /wp:group -->
 
-<!-- wp:template-part {"slug":"footer-writer","area":"footer","tagName":"footer"} /-->
+<!-- wp:template-part {"slug":"footer-writer","area":"footer","tagName":"footer","theme":"twentytwentyfour"} /-->
--- a/patterns/archive-writer.php
+++ b/patterns/archive-writer.php
@@ -2,7 +2,7 @@
 /**
  * Title: Writer Archive Template
  * Slug: twentytwentyfour/archive-writer
- * Template Types: archive
+ * Template Types: archive, page
  * Viewport width: 1400
  * Inserter: no
  */
--- a/patterns/archive-writer.php
+++ b/patterns/archive-writer.php
@@ -16,41 +16,13 @@
 
 <!-- wp:group {"tagName":"main","style":{"spacing":{"blockGap":"0","margin":{"top":"0"}}},"layout":{"type":"default"}} -->
 <main class="wp-block-group" style="margin-top:0">
-	<!-- wp:query {"query":{"perPage":3,"pages":0,"offset":"0","postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"align":"full","layout":{"type":"constrained"}} -->
-	<div class="wp-block-query alignfull">
-		<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50","left":"0","right":"0"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"default"}} -->
-		<div class="wp-block-group" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--50);padding-right:0;padding-bottom:var(--wp--preset--spacing--50);padding-left:0"><!-- wp:query-title {"type":"archive","align":"wide","style":{"typography":{"lineHeight":"1"},"spacing":{"margin":{"top":"0"}}},"fontSize":"x-large"} /-->
-			<!-- wp:spacer {"height":"var:preset|spacing|50","style":{"spacing":{"margin":{"top":"0","bottom":"0"}}}} -->
-			<div style="margin-top:0;margin-bottom:0;height:var(--wp--preset--spacing--50)" aria-hidden="true" class="wp-block-spacer"></div>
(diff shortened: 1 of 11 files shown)


MaggieCabrera:
Ok I think there's been a misunderstanding, also the figma is not correct (I think!).

- I saw your comment on figma, the footer is the one that we use for writer, that seems like an old version that never got updated
- I think we still need the search form in this, and that's shared between all 3 different search templates. 
- The query loop area should be shared between index, archive and search, is there any reason why that can't happen if they are inheriting the query from the page? this should be in a pattern so we don't have duplicated code to maintain

carolinan:
I thought there were going to be 3 search result pages, the default (template), writer, portfolio (patterns).

MaggieCabrera:
> I thought there were going to be 3 search result pages, the default (template), writer, portfolio (patterns).

yes, but each one of them shares the query loop with their respective index and archive pages. So you end up with something like:

Search portfolio pattern - has a search block and a portfolio-query-loop pattern
Archive portfolio pattern - has a query title and the same portfolio-query-loop pattern
Index portfolio pattern - has an h1 and the same portfolio-query-loop pattern

Then repeat for the other 2.

Is that clearer?

carolinan:
It may be easier to work with the templates if the file name started with writer instead?

patterns/posts-writer.php
```diff
@@ -0,0 +1,39 @@
+<?php
+/**
+ * Title: Posts for Writer templates
+ * Slug: twentytwentyfour/posts-writer
+ * Inserter: no
```
MaggieCabrera: if this is hidden, we need to prefix the file with hidden-

patterns/search-writer.php
```diff
@@ -0,0 +1,29 @@
...
+	<!-- wp:template-part {"slug":"header","theme":"twentytwentyfour"} /-->
+</div>
+<!-- /wp:group -->
+
+<!-- wp:group {"tagName":"main","style":{"spacing":{"blockGap":"0","margin":{"top":"0"}}},"layout":{"type":"default"}} -->
+<main class="wp-block-group" style="margin-top:0">
+	<!-- wp:group {"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group">
+		<!-- wp:query-title {"type":"search","style":{"typography":{"lineHeight":"1"},"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|30"}}}} /-->
+		<!-- wp:search {"label":"Search","showLabel":false,"buttonText":"Search","fontSize":"medium"} /-->
```
carolinan: This should probably have the same column as in https://github.com/WordPress/twentytwentyfour/pull/374

patterns/index-writer.php
```diff
@@ -0,0 +1,30 @@
+<?php
+/**
+ * Title: Writer Index Template
+ * Slug: twentytwentyfour/index-writer
+ * Template Types: index
```
MaggieCabrera: I think we need to add page and home here too

patterns/search-writer.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Title: Writer Search Results Template
+ * Slug: twentytwentyfour/search-writer
+ * Template Types: search
```
MaggieCabrera: this should have page too

patterns/archive-writer.php
```diff
@@ -10,47 +10,19 @@
```
MaggieCabrera: We probably should add page to the template types

MaggieCabrera:
@richtabor I'm having a hard time testing these kinds of patterns without downright altering the code and inserting the pattern like Carolina says in the description. Even trying to paste them inside a page to test is not working for me at all. How can we make them more discoverable?

MaggieCabrera:
I think the query loop pattern is perfectly fine to show to users as a variation for the query loop for both the writer and default templates

patterns/index-writer.php
carolinan: wp_template?

richtabor:
>  Even trying to paste them inside a page to test is not working for me at all. How can we make them more discoverable?

Just set them to be allowed within the inserter (for testing).

patterns/archive-writer.php
richtabor: Why would this archive template be suggested when you are creating a page template?

patterns/index-writer.php
MaggieCabrera: that would be Post Type instead of Template Type, maybe that's what we are missing, but I see other similar patterns don't have it? maybe those need it as well... I'm already confused with all the back and forth

richtabor:
We should make template pattern filenames/slugs prefixed with `template-` to help keep concepts organized. Like "template-home-portfolio.php" is.

richtabor:
> @richtabor I'm having a hard time testing these kinds of patterns 

We have `page` assigned as a templateType (which I think we should remove). But for now you can create a template for a page and see the template selection modal: 

CleanShot 2023-09-18 at 10 46 23

patterns/index-writer.php
carolinan: Me too

MaggieCabrera:
> We should make template pattern filenames/slugs prefixed with `template-` to help keep concepts organized. Like "template-home-portfolio.php" is.

I like this idea because it helps avoid conflicts with the template hierarchy

MaggieCabrera (approved):
I did some renaming so the pattern makes more sense to users and to the file names as Rich suggested, it's clearer if they are grouped together

MaggieCabrera:
I'm bringing this in and we can adapt the post/template types as we see fit once all template/patterns are in, I don't want to stall this more

PR #379, opened 2023-09-15
Use `esc_attr__` instead of `esc_attr` to make strings translatable.

Soean:
**Description**
This pull request changes the function calls from `esc_attr` to `esc_attr__` to make strings translatable.

`esc_attr( string $text )`: Escapes for HTML attributes and only accepts one parameter, the $text.

`esc_attr__( string $text, string $domain = 'default' )`: Retrieves the translation of $text and escapes it for safe use in an attribute.


MaggieCabrera:
Do you mind updating the readme file in the root too, please? Thank you!

Soean:
@MaggieCabrera Oh yes that makes sense, I updated the readme.

MaggieCabrera (approved):

PR #382, opened 2023-09-15
Add: Single portfolio template

MaggieCabrera:
**Description**

This PR introduces the template for Single portfolio designed in figma:

https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=492%3A2741&mode=dev

**Screenshots**

Screenshot 2023-09-15 at 17-47-36 Atque cupiditate sed et est qui – wp-stable

**Testing Instructions**

Create a post and select the template on the sidebar to be single portfolio.

### DIFF
--- a/patterns/template-single-portfolio.php
+++ b/patterns/template-single-portfolio.php
@@ -1,6 +1,16 @@
+<?php
+/**
+ * Title: Portfolio Single post Template
+ * Slug: twentytwentyfour/template-single-portfolio
+ * Template Types: posts, single
+ * Viewport width: 1400
+ * Inserter: no
+ */
+?>
+
 <!-- wp:group {"style":{"position":{"type":"sticky","top":"0px"}},"layout":{"type":"default"}} -->
 <div class="wp-block-group">
-	<!-- wp:template-part {"slug":"header"} /-->
+	<!-- wp:template-part {"slug":"header","theme":"twentytwentyfour"} /-->
 </div>
 <!-- /wp:group -->
 
--- a/templates/single-portfolio.html
+++ b/templates/single-portfolio.html
@@ -24,6 +24,11 @@
 	<div class="wp-block-column" style="flex-basis:55%"><!-- wp:post-content /--></div>
 	<!-- /wp:column --></div>
 	<!-- /wp:columns -->
+
+	<!-- wp:spacer {"height":"var:preset|spacing|40"} -->
+	<div style="height:var(--wp--preset--spacing--40)" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer -->
+
 </main>
 <!-- /wp:group -->
 
--- a/templates/single-portfolio.html
+++ b/templates/single-portfolio.html
@@ -9,15 +9,15 @@
 	<div style="height:var(--wp--preset--spacing--40)" aria-hidden="true" class="wp-block-spacer"></div>
 	<!-- /wp:spacer -->
 	
-	<!-- wp:post-featured-image {"align":"wide","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|50"}}}} /-->
+	<!-- wp:post-featured-image {"align":"wide","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|20"}}}} /-->
 	
 	<!-- wp:columns {"align":"wide"} -->
 	<div class="wp-block-columns alignwide"><!-- wp:column {"width":"45%"} -->
-	<div class="wp-block-column" style="flex-basis:45%"><!-- wp:post-title /--></div>
+	<div class="wp-block-column" style="flex-basis:45%"></div>
 	<!-- /wp:column -->
 	
 	<!-- wp:column {"width":"55%"} -->
-	<div class="wp-block-column" style="flex-basis:55%"><!-- wp:post-content /--></div>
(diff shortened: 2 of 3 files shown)


richtabor:
This should be abstracted as a pattern, like we did here https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/template-home-portfolio.php. Otherwise it’s confusing having the template available in the site editor (while it’s not used anywhere).

MaggieCabrera:
> This should be abstracted as a pattern, like we did here https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/template-home-portfolio.php. Otherwise it’s confusing having the template available in the site editor (while it’s not used anywhere).

I don't think I agree. It's fine if we decide we want this as a pattern, but the template for single and pages need to stay so people can use them for specific posts. Maybe we need to think of remaining these. Bea created the 3 concepts to showcase the theme as 3 different use cases, but the idea behind that was that theme could support multiple "post formats". If we only provide full template replacing, that defeats the purpose, I'd like to keep the theme flexible providing the ability to change between them.

richtabor:
> I don't think I agree. It's fine if we decide we want this as a pattern, but the template for single and pages need to stay so people can use them for specific posts. 

This should be an alternative single template type, not a standalone template. That's what we're doing with the home page (and other subsequent templates, right)? If we have a "Single Portfolio" template in the site editor, it looks like the actual single portfolio template for a custom post type.

richtabor:
> Create a post and select the template on the sidebar to be single portfolio.

I would not expect this to be a one-off template flow, but rather "I want to use this template for my single posts" (which supports the original idea of having three concepts to showcase the theme in three different ways.

MaggieCabrera:
> I would not expect this to be a one-off template flow, but rather "I want to use this template for my single posts" (which supports the original idea of having three concepts to showcase the theme in three different ways.

That's where I don't agree. I might be misunderstanding the whole concept behind the design, but I think the idea was to have 3 different concepts to showcase use cases for different types of content, because from a design perspective it's harder to find a use case that will need them all, but the goal is to give the flexibility depending on their own content. Sure some sites will just do fine with one of the single templates, but not all of them. This was a take on how post formats could be implemented on a block theme natively. If I have post with just an image, I will want to use the single-portfolio that shows it very prominently, but for my regular text based posts I will want to use single-writer to show my sidebar instead. 

Then again, I might be understanding this completely wrong but the way we set up the demo sites is just a way of of showing the users how they can set up their sites, but I don't think it's great that we are forcing them into just one way of displaying their content if they want to mix and match? Else it would have made sense to just make child themes instead

richtabor:
I don't think we should mix post formats and templates tbh.

carolinan:
I would expect to be able to create a post and select any from the template selection: default, writer, portfolio, sidebar
What would be more confusing to me is that they have different footers.

richtabor:
> I would expect to be able to create a post and select any from the template selection:

I don't think I agree. I expect it's confusing mixing post formats and templates here. 

> default, writer, portfolio, sidebar

I don't think these should be surfaced as such perhaps. 🤔

richtabor:
> What would be more confusing to me is that they have different footers.

Agreed.

richtabor:
If we don't make this as a pattern, it's not available as an alternate template within the site editor (for pages and for single). Makes it inconsistent with how we did template-home-portfolio.php. 

Related to https://github.com/WordPress/gutenberg/issues/54562

MaggieCabrera:
> I would expect to be able to create a post and select any from the template selection: default, writer, portfolio, sidebar What would be more confusing to me is that they have different footers.

You have a point about the footers. I'm not sure how we can fix that. 

@richtabor and I discussed this a bit more and came to the agreement that it's ok to have the sidebar/no sidebar single as templates, but this one is a bit more niche and confusing for users and should stay as a pattern like we have done for the alternative home pages. I will update this

MaggieCabrera:
@richtabor can you have a look and give this a 👍 so we can get the pattern in? once were have all of them we can discuss if we need changes to the template/post types for all these patterns

patterns/template-single-portfolio.php
```diff
@@ -0,0 +1,41 @@
...
+	<!-- /wp:columns -->
+
+	<!-- wp:spacer {"height":"var:preset|spacing|40"} -->
+	<div style="height:var(--wp--preset--spacing--40)" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer -->
+
+</main>
+<!-- /wp:group -->
+
+<!-- wp:template-part {"slug":"footer-portfolio","area":"footer","tagName":"footer"} /-->
```
luminuu, line 41: I think the `"theme":"twentytwentyfour"` is missing. I also have an issue with the footer not being properly rendered on inserting, but I'm not sure if this maybe just a local issue with my installation.

luminuu (changes requested):

patterns/template-single-portfolio.php
MaggieCabrera: https://github.com/WordPress/gutenberg/pull/54595 this is the issue related to that

luminuu: Should it be added or not in this case?

MaggieCabrera: we have not been consistent about it. I think that shouldn't block this PR because we will have to review every single pattern to add or remove this. I'm hoping this bug will be fixed upstream and we don't need the theme slug at all once that PR I linked gets merged

luminuu (approved):

PR #383, opened 2023-09-15
Changed Message Icon / Image URL in footer-portfolio.php

shail-mehta:
Changed Message Icon / Image URL in footer-portfolio.php

Use  `<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/icon-message.webp`


MaggieCabrera (approved):
thanks for catching this one!

PR #385, opened 2023-09-16
Fixed Pricing Pattern Separator Issue

shail-mehta:
Fixed Pricing Pattern Separator Issue. (Used is-style-wide for separator)

### **Before:** 
before

### **After:**
after


carolinan (approved):
This works well in my test, thank you.

PR #381, opened 2023-09-15
Sidebar: cleanup and enhancements

MaggieCabrera:
**Description**

This PR removes the sidebar-writer pattern because it's virtually the same as this one but with french text. I changed a few blocks:

- The image is now the avatar
- The post meta uses the block pattern
- The useful links use the nav block instead of another query loop

**Screenshots**

Before:

Screenshot 2023-09-15 at 17 13 51

After:

Screenshot 2023-09-15 at 17 13 09

**Testing Instructions**

Create a post and select single with sidebar as the template. Make sure to add a description to the author page in your wp install to see all the content

PR #390, opened 2023-09-17
Fixed Separator Width Issue in Sidebar

shail-mehta:
## **Fixed Separator Width Issue in Sidebar  (Used is-style-wide for separator)**

**Before:** 
before-separator

**After:** 
after-separator


MaggieCabrera:
I'll add this changes to https://github.com/WordPress/twentytwentyfour/pull/381 instead, thank you!

PR #387, opened 2023-09-17
Fixed Spacing issue in Feature With Images Pattern

shail-mehta:
### **Before Ipad View:** 
ipad-before

### **After Ipad View:** 
ipad-after

### **Before Mobile View:** 
mobile-before

### **After Mobile View:** 
mobile-after

PR #391, opened 2023-09-17
Another round of pattern improvements

richtabor:
**Description**

Cleans up the patterns we did not add presets to, uses proper presets where they weren't applied, improves mobile, adds viewport width where necessary. 

CleanShot 2023-09-17 at 14 20 59

CleanShot 2023-09-17 at 14 21 10
CleanShot 2023-09-17 at 14 21 23

Additional props: [shail-mehta](https://github.com/shail-mehta/shail-mehta) (related: https://github.com/WordPress/twentytwentyfour/pull/387)


beafialho:
Can the RSVP vertical text be turned the other way like the design? Also, it doesn't have the hand symbol. Does it need to be an image?

Captura de ecrã 2023-09-17, às 19 37 00

Services pattern has the wrong image, it seems, from your screenshots:

Captura de ecrã 2023-09-17, às 19 37 42

richtabor:
> Services pattern has the wrong image, it seems, from your screenshots:

Yes, we need to consolidate some of the images to reduce the load. This one was only used once, so I applied that other abstract image for both. 

> Can the RSVP vertical text be turned the other way like the design? 

Yes! When the next Gutenberg release lands, we should be able to adjust this. 

> Also, it doesn't have the hand symbol. Does it need to be an image?

It probably will need to be add another image. I’m a bit hesitant to add another, especially if it’s only used once. :/

richtabor:
Alternatively, if the images are available to direct link, we could consider that (instead of bundling images in the assets directory). Not a blocker though.

carolinan:
The theme can not have remote images. Remote files can be replaced and removed at any time so it is not allowed in the theme directory. This is because people abused it in the past switching their media to adverts.

**What about media resources that .org controls?**
On request by the meta team, other theme authors are not allowed to "hotlink" images from openverse, photo directory, or  other resources uploaded to .org,  so the default theme can not use that, both for fairness and to not set an example that others can't copy.

We have also seen images used in the pattern directory break in the past, both from being removed from the source but also because that image provider does not allow hotlinking.

MaggieCabrera:
The about template can't be removed because it's used in the home writer pattern/template alternative and needs translation

MaggieCabrera:
> The about template can't be removed because it's used in the home writer pattern/template alternative and needs translation

Never mind, I see what you did is actually rename it, we just needed to update the slug on the home writer pattern

MaggieCabrera (approved):
This looks good, I'm bringing this in before we create more conflicts

PR #393, opened 2023-09-18
Add rust theme variation

richtabor:
**Description**
Seeing how a variation without a lot of accent colors will look and feel. 

One thing that stands out to me is how tied the images are to the default variation. An idea we can potentially try is to have a duotone preset on images within patterns that does nothing on the default variation, but alternate variations could style. 

But I don't think this is a blocker to this (and subsequent variations). Rather something we can explore. 

For example, this pattern has a black and white duotone (`dutone-1`) applied, and the Rust variation here makes duotone-1 rust to beige. That makes this image fit in better with the new variation, without having to apply the duotone to the image block globally (which I don't think is a good idea as it is disorienting having it auto-applied). 

CleanShot 2023-09-17 at 21 28 55
CleanShot 2023-09-17 at 21 28 27

### Pattern visuals

CleanShot 2023-09-17 at 21 24 02
CleanShot 2023-09-17 at 21 26 20

### DIFF
--- a/styles/rust.json
+++ b/styles/rust.json
@@ -1,0 +1,95 @@
+{
+	"version": 2,
+	"title": "Rust",
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"settings": {
+		"color": {
+			"duotone": [
+				{
+					"colors": [
+						"#A62B0C",
+						"#F3F0E7"
+					],
+					"slug": "duotone-1",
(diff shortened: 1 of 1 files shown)


huzaifaalmesbah (commented):
I checked your Rust Variation. Looks good for me.

rust theme variation home page

huzaifaalmesbah (approved):

PR #398, opened 2023-09-18
Update search form styling

### LINKED ISSUE #389: Improve search form appearance
Unfortunately there's no UI to perfect these forms accurately but can we improve this? It's looking very unpolished. The placeholder font weight should be **Regular** and the input field border radius should be the same as the theme's buttons.

These details make a difference.

| Reality      | Expectation |
| ----------- | ----------- |
| Captura de ecrã 2023-09-17, às 12 37 57   |  Captura de ecrã 2023-09-17, às 12 41 24    |

carolinan:
**Description**
Updates the default border radius for the button and input field.
Removes the font weight.

Closes https://github.com/WordPress/twentytwentyfour/issues/389

**Screenshots**

**Testing Instructions**
Add a search block with different settings to confirm that the design is correct ( see the issue ).

In the block's sidebar settings, adjust the border radius and confirm that the user's settings still overrides the default.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -459,6 +459,9 @@
 					"fontSize": "var(--wp--preset--font-size--small)"
 				}
 			},
+			"core/post-comments-form": {
+				"css": "& textarea, input{border-radius:.33rem}"
+			},
 			"core/comments-pagination": {
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--small)"
@@ -506,6 +509,9 @@
 						"left": "var(--wp--preset--spacing--10)"
 					}
 				}
+			},
+			"core/loginout": {
+				"css": "& input{border-radius:.33rem;padding:calc(0.667em + 2px);border:1px solid #949494;}"
 			},
 			"core/navigation": {
 				"elements": {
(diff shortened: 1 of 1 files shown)


theme.json
```diff
@@ -716,9 +716,16 @@
...
+				"css": "& .wp-block-search__input{border-radius:.33rem}",
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--small)",
-					"fontWeight": "600"
+					"fontSize": "var(--wp--preset--font-size--small)"
+				},
+				"elements": {
+					"button": {
+						"border": {
+							"radius": { "ref": "styles.elements.button.border.radius" }
```
carolinan, line 732: This will keep the radius up to date if we ever decide to change the button border radius.
Note that the CSS field does not work the same way, there the radius is hardcoded.

carolinan: **This is probably an editor bug**,  The border radius **_should_** already be inherited from the element button styles even without this code.
But I noticed that without this code, it only had the border radius on the front, not in the editor.

carolinan:
Login/out and comment form:
I added a small amount of CSS to theme.json. If we must re-style the login form button, it will take more CSS and it may be better to add that to style.css.
forms with input fields with border radius

MaggieCabrera (approved):

Issue #9, opened 2023-08-24
Templates


luminuu:
We need to add standard templates to the theme.

Pattern | Done | Reviewed | PR (or issue)
--- | --- | --- | ---  
Home | ✅  | 🔳 | N/A
Home Writer (pattern) | ✅ | 🔳 | N/A
Home Portfolio (pattern) | ✅ | 🔳 |  N/A
Single | ✅ | 🔳 |  N/A
Single Portfolio | ✅ | 🔳 | N/A
Single with Sidebar | ✅ | 🔳 | N/A
Page | ✅ | 🔳 | N/A
Page with Sidebar | ✅ | 🔳 | N/A
Page Portfolio  | ✅ | 🔳 | N/A
Page no title | ✅ | 🔳 | N/A
Index | ✅ | 🔳 | N/A
Index Writer (pattern) | ✅ | 🔳 | N/A
Index Portfolio (pattern) | 🔳 | 🔳 | #425 (issue)
Archive | ✅ | 🔳 | N/A
Archive Writer (pattern) | ✅ | 🔳 | N/A
Archive Portfolio (pattern) | 🔳 | 🔳 | #368
Search | ✅ | 🔳 | N/A
Search Portfolio (pattern) | 🔳 | 🔳 | #426 (issue)
404 | ✅ | 🔳 | N/A

How can we achieve different default templates for each of the subjects (Entrepreneur/Business, Writer/Blogger, Photographer/Creative)?

MaggieCabrera:
I updated this looking at the figma grid of templates. Some of the templates are going to be patterns and need to be turned into such and removed from the template list. This should be our priority for Beta 1 @luminuu

MaggieCabrera:
The reviewed column alludes to a design review after the PRs are merged

carolinan:
@beafialho commented on Figma that the writer search template is only for the demo. 
In this issue for the _writer_ I am asked to add all 3 patterns for the _portfolio_. https://github.com/WordPress/twentytwentyfour/pull/377#issuecomment-1721455593
This is confusing and time confusing. 
Please label all templates clearly.

beafialho:
> @beafialho commented on Figma that the writer search template is only for the demo.
In this issue for the writer I am asked to add all 3 patterns for the portfolio. https://github.com/WordPress/twentytwentyfour/pull/377#issuecomment-1721455593
This is confusing and time confusing.
Please label all templates clearly.

[...]

PR #399, opened 2023-09-18
Add page with sidebar

carolinan:
**Description**

Re-adds the page with sidebar template.
This template and the sidebar pattern will need several iterations to spacing and to fix accessibility issues.

Only the "writer" design has a page with sidebar so I am not sure that this is correct.
There is a page template with sidebar listed here: https://github.com/WordPress/twentytwentyfour/issues/9

Figma: https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?node-id=511%3A1757&mode=dev

**Screenshots**

**Testing Instructions**

Select or create a new page.
In the block editor, in the Page panel, select template and choose "Page with Sidebar"
Save.
View the page on the front.

### DIFF
--- a/templates/page-with-sidebar.html
+++ b/templates/page-with-sidebar.html
@@ -4,10 +4,11 @@
 </div>
 <!-- /wp:group -->
 
-<!-- wp:group {"style":{"spacing":{"margin":{"top":"0vh","bottom":"0vh"},"padding":{"top":"10vh","bottom":"8vh"}}},"layout":{"type":"constrained"}} -->
-<div class="wp-block-group" style="margin-top:0vh;margin-bottom:0vh;padding-top:10vh;padding-bottom:8vh">
-	<!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"top":"1rem","left":"1rem"}}}} -->
-	<div class="wp-block-columns alignwide">
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group">
+
+	<!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"top":"1rem","left":"1rem"},"padding":{"top":"var:preset|spacing|50"},"margin":{"bottom":"var:preset|spacing|40"}}}} -->
+	<div class="wp-block-columns alignwide" style="margin-bottom:var(--wp--preset--spacing--40);padding-top:var(--wp--preset--spacing--50)">
 		<!-- wp:column {"width":"10%"} -->
 		<div class="wp-block-column" style="flex-basis:10%"></div>
 		<!-- /wp:column -->
--- a/templates/page-with-sidebar.html
+++ b/templates/page-with-sidebar.html
@@ -1,0 +1,52 @@
+<!-- wp:group {"style":{"position":{"type":"sticky","top":"0px"}},"layout":{"type":"default"}} -->
+<div class="wp-block-group">
+	<!-- wp:template-part {"slug":"header"} /-->
+</div>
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
Like I said somewhere else (I totally lost where), we probably should discuss the naming of some of these templates. They make sense to me for things like the home page, where the concept of portfolio/writer/business is more apparent because of the content we've chosen to showcase. But when it comes to page/single, we are talking about the user's content, and to me it feels much less confusing and informative to be descriptive about what the template looks like. In this case Page with sidebar conveys much more to me than Page Writer, and it let's me know easily when I'm creating a page if this is the template I need to switch to when I want my page to have just that, a sidebar. Opinions? @richtabor @beafialho

carolinan:
Maybe it should just be a neutral page pattern without the writer footer?

MaggieCabrera:
> Maybe it should just be a neutral page pattern without the writer footer?

good question

richtabor:
Yes, I'm thinking headers and footers should be consistent across templates — that way if replaced in the template part flow, the changes persist across templates. Otherwise you'd have multiple headers, unintentionally.

beafialho:
> In this case Page with sidebar conveys much more to me than Page Writer

Agree on the "Page With Sidebar" conveying more, as it can be used well throughout all three cases, but for the others (Archives, Search, Posts) Entrepreneur, Photographer and Writer make sense because they're descriptive and suggestive. 

If I read "Photographer Archive", I imagine it's designed to be image focused, and if I read "Writer Archive" I assume it's designed to feature mostly text. Another advantage of naming them similarly is it becomes a hint that if you use the "Photographer Archive" template with the "Photographer Footer", your site will probably look more cohesive, and so on.

That said though, if we agree on naming them all more literally, I'm open to suggestions.

templates/page-with-sidebar.html
```diff
@@ -0,0 +1,53 @@
...
+	</div>
+	<!-- /wp:column -->
+
+	<!-- wp:column {"width":"10%"} -->
+	<div class="wp-block-column" style="flex-basis:10%"></div>
+	<!-- /wp:column -->
+
+	<!-- wp:column {"width":"30%"} -->
+	<div class="wp-block-column" style="flex-basis:30%">
+		<!-- wp:template-part {"slug":"sidebar"} /-->
```
richtabor, line 41: Nice! https://github.com/WordPress/twentytwentyfour/issues/408

richtabor:
Why not just use the existing template for posts and add `post` to the postTypes? We don't need a whole separate template for these if they do the same thing. 


```
		{
			"name": "sidebar",
			"postTypes": [
				"page",
				"post"
			],
			"title": "Sidebar"
		}
```
CleanShot 2023-09-18 at 15 46 30

CleanShot 2023-09-18 at 15 46 06

richtabor (dismissed):
Why not just use the existing template for posts and add `post` to the postTypes? We don't need a whole separate template for these if they do the same thing. 


```
		{
			"name": "sidebar",
			"postTypes": [
				"page",
				"post"
			],
			"title": "Sidebar"
		}
```
CleanShot 2023-09-18 at 15 46 30

CleanShot 2023-09-18 at 15 46 06

carolinan:
Because the page templates don't display the post meta, next and previous post, and has no comment form.

richtabor:
Good point. Bummer we have nearly duplicative templates. :/

carolinan:
Yeah its a shame we can't use more conditions.

luminuu (approved):
LGTM!

Issue #363, opened 2023-09-13
`border-radius` consistency across all patterns


colorful-tones:
**Description**

It is recommended we set `settings.custom.border.radius` to control a consistent `border-radius` value throughout patterns. Right now, I'm seeing a lot of `"border":{"radius":"16px"}` in patterns. So, maybe we go with `16px`? Either way, it should reference a variable.

We could also consider utilizing an existing spacing preset. Like @richtabor did in [`c7f123d`](https://github.com/WordPress/twentytwentyfour/commit/c7f123dc9b2e0fda8fde5a7e82198e0ebf692ba7)

Also, someone should run through the `theme.json` and look to streamline values to variable.

Here is a list of affected patterns that would need to be updated:

- [ ] `centered-statement.php`
- [ ] `clients.php`
- [ ] `cta.php`
- [ ] `pricing.php`

Maybe `testimonial.php` as well? I'm seeing `"border":{"radius":"999px"}`

richtabor:
> Right now, I'm seeing a lot of "border":{"radius":"16px"} in patterns.

If it's a block that does not support rounded, that's what's used. Perhaps we add block style variations for those few block, like image had "rounded"?

MaggieCabrera:
> > Right now, I'm seeing a lot of "border":{"radius":"16px"} in patterns.
> 
> If it's a block that does not support rounded, that's what's used. Perhaps we add block style variations for those few block, like image had "rounded"?

I think that makes it more consistent, yes

MaggieCabrera:
We changed this so we are using a preset variable so I'm closing this one

PR #401, opened 2023-09-18
Try rounded group block style

carolinan:
**Description**
border-radius consistency across all patterns by adding a rounded corner block style for groups.
For https://github.com/WordPress/twentytwentyfour/issues/363

**Screenshots**
group with border radius

**Testing Instructions**
Add a group block, and select the block "rounded" block style.


carolinan:
Opinion: This does not work well without padding and background color.

MaggieCabrera:
I agree with that opinion. Even more so because it's a variation. Let's use the white color from the palette for the background, we've used that style in some of our patterns.

carolinan:
The left and right padding is being overridden by:

```
has-global-padding :where(.has-global-padding:not(.wp-block-block)) {
	padding-right: 0;
	padding-left: 0;
}
```

styled group block is missing left and right padding

carolinan:
I don't think we should use important for the padding to override the styles from core, so I will close this.

PR #403, opened 2023-09-18
Removed themename in template parts

shail-mehta:
Removed `"theme":"twentytwentyfour"` in Template Parts as per guideline

**Files:**
patterns/template-home-portfolio.php
patterns/template-home-writer.php

### DIFF
--- a/patterns/hidden-sidebar.php
+++ b/patterns/hidden-sidebar.php
@@ -38,7 +38,7 @@
 	<div class="wp-block-query"><!-- wp:post-template -->
 	<!-- wp:group {"style":{"spacing":{"blockGap":"2px"}},"layout":{"type":"flex","orientation":"vertical"}} -->
 	<div class="wp-block-group"><!-- wp:post-title {"isLink":true,"fontSize":"medium"} /-->
-	<!-- wp:template-part {"slug":"post-meta","theme":"twentytwentyfour"} /-->
+	<!-- wp:template-part {"slug":"post-meta"} /-->
 	</div>
 	<!-- /wp:group -->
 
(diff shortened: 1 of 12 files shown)


richtabor (dismissed):
This is blocks at the moment. 

The `theme` is currently required to include template parts _within_ patterns, like we're doing in these two files. Otherwise they don't render: 

CleanShot 2023-09-18 at 12 48 08

Related: https://github.com/WordPress/gutenberg/issues/53194#issuecomment-1723598009

MaggieCabrera:
I pinged @scruffian on this issue. I wonder, do the template parts still work with the theme slug present if the theme is installed in a subfolder/the folder has a different name than twentytwentyfour?

MaggieCabrera:
https://github.com/WordPress/gutenberg/pull/54595 has been merged and I would like to merge this by end of day today so it makes it to the first build in beta 1. This ensures that template parts will show no matter the folder your theme is installed in.

Right now to test it we need to use the latest GB trunk. I've marked the PR for backport so it needs to make it to the final GB release before beta for this to work without the plugin installed.

Lets test that this is working as intended.

scruffian (approved):
LGTM

MaggieCabrera:
When is it best to merge this? @desrosj if I include this change on the first build of the theme at the current state, the template parts will fail. I'm assuming https://github.com/WordPress/gutenberg/pull/54595 will get backported during Tuesday, so maybe it's best if we sit on this and add it to that build last minute?

desrosj:
Let's proceed with the original PR and then create a follow up one for then. If there are any other changes that are ready, we can include them as well.

MaggieCabrera:
We can bring this in now, I tested with core trunk and we don't need the theme slugs anymore

PR #406, opened 2023-09-18
Use an alternate "portfolio" template for pages/posts

richtabor:
**Description**

@MaggieCabrera Explored your PR as template type pattern, instead of a template within the site editor sidebar. With this method, you can assign the template to pages or posts (holistically, or single). 

**Screenshots**

https://github.com/WordPress/twentytwentyfour/assets/1813435/59fae4fc-d687-4f6b-b74d-6efde044be63


carolinan:
This pattern selection only works if the template is blank. By making it a template, any post and page be changed without having to delete anything.

It creates a new template in the database, and by having the template as part of the theme, the user doesn't _have_ to do this step.

richtabor:
> any post and page be changed without having to delete anything.

Currently you can do this with posts and pages like in my screen recording, without having to delete anything. 

> by having the template as part of the theme, the user doesn't have to do this step.

Yes, but the consequence is that it's confusing to see a Portfolio template in the site editor, if you don't have a portfolio site. 

We need to decide on the route we go: 
- Do we make templates available everywhere as a one-off application (essentially a post format type deal), or...
- Do we leverage template types so that _alternates_ are available to swap out holistically in the site editor?

I lean towards the latter, where we register alternate templates, not make all templates always available. Otherwise it's confusing seeing alternate templates in the site editor and trying to figure out how to make one of those your page template too. 

The same argument holds for template parts (header/footer) where we don't want them all always visible, but you can swap to an alternate and have it consistent throughout your site (https://github.com/WordPress/twentytwentyfour/issues/9#issuecomment-1724150856).

carolinan:
Creating a new template is not at all intuitive when what you want it so change the template.
Custom page and post templates have been available in themes, including default themes, for many more years than the Site Editor has been available.

richtabor:
> Creating a new template is not at all intuitive when what you want it so change the template.

I'm thinking at the _site_ level, not the page level. Do you want your single posts to look like A, or B? Not every time you want to publish a post, you decide if you want A or B. 

> Custom page and post templates have been available in themes, including default themes, for many more years than the Site Editor has been available.

I don't understand the relevance.

carolinan:
The relevance is that it is easy for current users to change templates in the block editor. By making it a pattern only, that option is not available.

Site level and post level does not have to be mutually exclusive, not one or the other.

richtabor:
I would rather _not_ include a bunch of templates that confuse the site editor experience. As-is, the list of templates is very confusing, and most are unlikely to be relevant to the majority of sites.

carolinan:
That rebase did NOT go as expected, sorry Rich. I don't know how to revert it.

carolinan:
I recorded some videos of template swapping, but then realised that the page template for the portfolio was never removed, only the posts template. So my videos still used the html template not the pattern.

MaggieCabrera:
I think the problem with the rebase is that this branch wasn't created from trunk. Do you mind if we do this on my PR instead? 

I agree that polluting the list of templates with confusing options is probably not the best idea. A portfolio site would swap to this particular pattern as the default page/post template and then they have the option to select one of the other templates for their text-based posts instead. We need to iron out and document the different flows because it's hard to follow for us who are working on this, so I expect it to be confusing for users anyway. Also, there's work being done on these flows that even if they don't make it to 6.4, they most likely will for 6.5 which is still early in the lifecycle of TT4 and I'm happy with that as long as they get there.

patterns/template-single-portfolio.php
```diff
@@ -0,0 +1,43 @@
+<?php
+/**
+ * Title: Portfolio Single Template
+ * Slug: twentytwentyfour/template-single-portfolio
+ * Template Types: page, single
```
MaggieCabrera, line 5: My PR ended up being mostly the same but I have posts/single here. I don't think we should have page here, maybe we should have all 3? we can revisit that

MaggieCabrera:
I'm closing this one since my PR was merged but they ended up being mostly the same implementation. Thank you for this

PR #413, opened 2023-09-18
Styles Variation Schema Move To Top

huzaifaalmesbah:
I follow previews block theme like TT3 and also WordPress developer guide  [Reference](https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-json/#developing-with-theme-json)
To use the schema at the beginning of json file


luminuu (approved):
Good catch!

PR #415, opened 2023-09-19
Stylesheet: CSS coding standards fixes, comment editing

sabernhardt:
- Adds spaces following a colon
- Adds spaces around the child combinator (`>`)
- Adds a leading zero for decimal values
- Uses double quotes for Arrow icon `list-style-type`
- Uses `background-color` for both Pill link background colors (hover state had `background`)
- Makes the spacing below comments consistent
- Edits comment text

### DIFF
--- a/style.css
+++ b/style.css
@@ -15,11 +15,12 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 */
 
 /*
- * Control the hover stylings of outline block style.
+ * Control the hover stylings of Button block's Outline style.
  * Unnecessary once block styles are configurable via theme.json
  * https://github.com/WordPress/gutenberg/issues/42794
  */
-.wp-block-button.is-style-outline>.wp-block-button__link:not(.has-background):hover {
+
+.wp-block-button.is-style-outline > .wp-block-button__link:not(.has-background):hover {
 	background-color: var(--wp--preset--color--contrast-2);
 	color: var(--wp--preset--color--base);
 	border-color: var(--wp--preset--color--contrast-2);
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -29,13 +30,14 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
  * Link styles
  * https://github.com/WordPress/gutenberg/issues/42319
  */
+
 a {
-	text-decoration-thickness: .0625em !important;
-	text-underline-offset: .15em;
+	text-decoration-thickness: 0.0625em !important;
```
sabernhardt: Why does this need `!important`? The linked Gutenberg issue is about setting custom values for underline styles such as `text-decoration-thickness`.

style.css
```diff
@@ -15,11 +15,12 @@ Tags: one-column, custom-colors, custom-menu, custom-logo, editor-style, feature
 */
 
 /*
- * Control the hover stylings of outline block style.
+ * Control the hover stylings of Button block's Outline style.
  * Unnecessary once block styles are configurable via theme.json
  * https://github.com/WordPress/gutenberg/issues/42794
  */
-.wp-block-button.is-style-outline>.wp-block-button__link:not(.has-background):hover {
+
```
sabernhardt, line 22: I added a space below some of the comments for consistency. It possibly could be removed from all of them instead.

style.css
luminuu: In favour of not having any thing with `!important` in the CSS, this can be removed. It does not fix anything.

sabernhardt: This PR removes the `!important` now.

luminuu (approved):
LGTM!

PR #416, opened 2023-09-19
Update footer-portfolio.php: Translations, Accessibility (use headings), and copyright

carolinan:
**Description**
In this PR I am **proposing** the following changes:

### Accessibility:

1. Remove the alt text from the envelope, so that screen reader users can skip this decorative image.
2. Turn 3 out of five paragraphs into headings. The larger text is a heading that brings context to the footer area, and was already styled as a heading. The two column headings gives context to the contact link and social links.

### Translation:

1. Adds missing translation functions.
2. Replaces the hardcoded theme name with the site title.  Here, I tried to use 1ch as the block spacing between the year and the site title, but this broke in the Site Editor. I used 6px instead but this can of course be adjusted.

_I also adjusted the spacing in the file, so it looks like there are more changes._

**Screenshots**
In the screenshot, my site title is "64" :)
alternative footer template part

Mobile:
Screenshot 2023-09-19 at 10 25 39

**Testing Instructions**
Add the footer portfolio template part to any template and test it in the editor and front.


carolinan:
Ok the site title row still needs work 😆 
broken mobile footer

Unless we remove it? This is the only footer that has copyright.

carolinan:
Better?
footer row with wrap instead of nowrap

luminuu:
Either the last one or would it be possible to wrap the copyright / designed with WP line just like the other content above it?

carolinan:
https://github.com/WordPress/twentytwentyfour/assets/7422055/cc85bc70-9e6f-4ec4-98c5-2c7b77fde9c1

MaggieCabrera (approved):

PR #422, opened 2023-09-19
Adding a 'name' attribute to the Cardo font family definitions. 

pbking:
**Description**

With the latest Gutenberg release a 'name' attribute is required in font family definitions.  I believe it has always been required, but the resources would still render.  Now, if it is missing the `<style id='wp-fonts-local'...` resource does not render.

**Testing Instructions**

Ensure Gutenberg plugin is latest (/trunk) and active.
Render any view with the default 'Cardo' font applied.
Note that the font face resources are used.

image

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -251,6 +251,7 @@
 						}
 					],
 					"fontFamily": "Cardo",
+					"name": "Cardo",
 					"slug": "heading"
 				}
 			],


MaggieCabrera (approved):

MaggieCabrera:
Thank you Jason!

PR #429, opened 2023-09-20
Template cleanup

### LINKED ISSUE #412: Remove Page Writer template
It's so close to the existing "Page" template, as the only change is the text alignment. 

I propose we remove this custom template and consolidate. It could be a pattern template, but it doesn't really need to be with such a small change. 

Removing this will reduce the noise in the site editor > templates view.

CleanShot 2023-09-18 at 15 19 21
CleanShot 2023-09-18 at 15 19 13

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/412

This PR removes Page writer, renames Page portfolio and removes the non existant templates from theme.json

Page writer looks almost exactly like page, the only difference being the alignment of post title so I don't think that merits having a full template. We can manually do that change for the demo site if we want to.


luminuu (approved):
LGTM

PR #430, opened 2023-09-20
Make post meta a template part

### LINKED ISSUE #409: Post meta should be referenced as template part, not the hidden pattern
If the post meta display was using a template part, then it can be reused throughout templates — and edited from one source. Otherwise, you'd have to make a post meta change to every template. 

I propose we make a template part for the post meta, then use that throughout, instead of referencing the pattern directly.

CleanShot 2023-09-18 at 13 35 54

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/409

This makes post meta a template part so it stays synced between templates if the user decides to change it

### DIFF
--- a/parts/post-meta.html
+++ b/parts/post-meta.html
@@ -1,0 +1,1 @@
+<!-- wp:pattern {"slug":"twentytwentyfour/post-meta"} /-->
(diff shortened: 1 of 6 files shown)


luminuu (approved):
LGTM

PR #431, opened 2023-09-20
Update README with more info on translations

MaggieCabrera:
**Description**

Adds translation info as per the suggestions in https://github.com/WordPress/twentytwentyfour/pull/334


luminuu (approved):
👍

PR #433, opened 2023-09-20
Style Variation: Ice

MaggieCabrera:
**Description**

This adds the Sans serif variation following the discussion from https://github.com/WordPress/twentytwentyfour/pull/367#issuecomment-1721272112

I wasn't sure about the font weights to use (also google fonts doesn't provide any italics for Inter). If this is good we can do the same for the dark sans serif one, which probably will need a name too.

Seems like the patterns are working as intended:

Screenshot 2023-09-20 at 11 18 25


beafialho:
Thanks @MaggieCabrera, I reviewed it and in the sans-serif variation, the headings have slightly smaller sizes. I set them in Global Styles and also set them to use Bold. Here's the variation file:

[ice_1.json.zip](https://github.com/WordPress/twentytwentyfour/files/12671732/ice_1.json.zip)

In Google Fonts, Inter doesn't provide italics, but it does [here](https://github.com/rsms/inter/releases/tag/v3.19), maybe we can use them?

MaggieCabrera:
Looks like I misundertood and Rich meant to use Inter on the body font, I will be updating with that change and adding the italics

MaggieCabrera:
@beafialho I'm sorry for the confusion but I misunderstood which font was going to change to Inter. Rich made the change to change the body font in the theme to Inter, so this variation the only change that is making is change the heading font to use System font. Does that sound ok to you? do you think it needs any other tweaks?

beafialho:
> Rich made the change to change the body font in the theme to Inter, so this variation the only change that is making is change the heading font to use System font. Does that sound ok to you? do you think it needs any other tweaks?

That sounds ok, it's looking good 👍🏼 In the file below are some tweaks: I made the headings **Bold**, and customised the quote and pullquotes.

[ice-2.json.zip](https://github.com/WordPress/twentytwentyfour/files/12701914/ice-2.json.zip)

MaggieCabrera:
there is another version designed on figma that is exactly like this one but with the colors of onyx, we still haven't build that. What should we name it?

luminuu (approved):

PR #434, opened 2023-09-20
Performance improvement: Ensure styles for the button block only get loaded when the button block is rendered

aristath:
**Description**

This is a performance improvement to ensure that styles for the button block only get enqueued when the block actually exists on the page.

**Testing Instructions**

Add a button block, set the style variation to outline, and check that the styles get applied properly,.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -65,3 +65,4 @@
 | @labunchemjong | @labunchemjong |
 | @thetwopct | @bonkerz |
 | @karmacharya50 | @karmacharya50 |
+| aristath | @aristath |
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -65,4 +65,4 @@
 | @labunchemjong | @labunchemjong |
 | @thetwopct | @bonkerz |
 | @karmacharya50 | @karmacharya50 |
-| aristath | @aristath |
+| @aristath | @aristath |
--- a/assets/css/button-outline.css
+++ b/assets/css/button-outline.css
@@ -1,0 +1,6 @@
+.wp-block-button.is-style-outline
+	> .wp-block-button__link:not(.has-background):hover {
+	background-color: var(--wp--preset--color--contrast-2);
+	color: var(--wp--preset--color--base);
+	border-color: var(--wp--preset--color--contrast-2);
+}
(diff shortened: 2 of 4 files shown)


CONTRIBUTORS.md
```diff
@@ -65,3 +65,4 @@
 | @labunchemjong | @labunchemjong |
 | @thetwopct | @bonkerz |
 | @karmacharya50 | @karmacharya50 |
+| aristath | @aristath |
```
MaggieCabrera:
```suggestion
| @aristath | @aristath |
```

MaggieCabrera:
Mmmh, we have more block styles. Should we do this for all of them? Will this do the opposite if a page uses all of them? (some full page patterns might)

MaggieCabrera:
In fact, is it more performant if we just add the CSS inline inside functions.php? It might be uglier, but the block styles don't usually have too many lines.

aristath:
> Mmmh, we have more block styles. Should we do this for all of them? Will this do the opposite if a page uses all of them? (some full page patterns might)

Ideally yes, we should be doing this for all blocks that have custom CSS.
Even if we end up adding custom CSS to _all_ blocks, this method will still generate fewer styles on the frontend, resulting in a faster (and more sustainable) website 👍

luminuu:
I'd lean towards Ari's suggestion, but if there are many more block styles, I'm not sure what the best practice is there. Can we get an overview how many block styles we have?

aristath:
> In fact, is it more performant if we just add the CSS inline inside functions.php? It might be uglier, but the block styles don't usually have too many lines.

The method in this PR will automatically handle inlining _small_ stylesheets.
The key to WP handling this automatically is the addition of the `path` data using the `wp_style_add_data()` function (see [this post on make for details](https://make.wordpress.org/core/2021/07/01/block-styles-loading-enhancements-in-wordpress-5-8/#inlining-small-assets))

MaggieCabrera:
> The method in this PR will automatically handle inlining _small_ stylesheets. The key to WP handling this automatically is the addition of the `path` data using the `wp_style_add_data()` function (see [this post on make for details](https://make.wordpress.org/core/2021/07/01/block-styles-loading-enhancements-in-wordpress-5-8/#inlining-small-assets))


Ooooooh, it does it for you? TIL! so smart. Thank you for teaching us this, I suggest we make a comment with a link to that make post for user's reference, see if we can spread the knowledge beyond just this repo, do you agree?

aristath:
>  suggest we make a comment with a link to that make post for user's reference, see if we can spread the knowledge beyond just this repo, do you agree?

Absolutely! Added some comments in the code 👍

MaggieCabrera (approved):

carolinan:
This is a great option for themes that use functions.php.

PR #441, opened 2023-09-20
Added font license

MaggieCabrera:
**Description**

Before we go ahead with the rest of the variations, let's reorganize our assets so they include the license files like TT2 does, so it serves as a reference for the other variations


luminuu (approved):
👍

PR #448, opened 2023-09-20
Asterisk block style for separator block

colorful-tones:
**Description**

An alternative to #51 and #365 . This allows the user to choose colors _and_ alignment. All alignments are centered for .alignfull, .alignwide, .aligncenter. If there is no alignment then the asterisk aligns left.

Screenshot 2023-09-20 at 1 37 28 PM


carolinan:
It can't be an HR: 
- Doesn't have any screen reader announcement, as it's a decorative element, not semantic

> An `hr` element is announced by screen readers, so I think that's only viable if the asterisk is a legitimate representation of a section break. Most screen readers announce it as 'separator'; VoiceOver announces it as 'horizontal splitter' or 'horizontal separator' depending on browser. So it would still be fairly verbose. The announcement is meaningful, however, as long as the semantics match the usage.

carolinan:
Eh I meant to comment not close

colorful-tones:
> It can't be an HR

🤦 I missed that.

> Eh I meant to comment not close

No worries!

MaggieCabrera:
the reason being it's being announced by screen readers and it's not semantically correct

MaggieCabrera:
also, does it work as an external image to change colors in chrome? last time I tried it only worked on firefox

colorful-tones:
> also, does it work as an external image to change colors in chrome? last time I tried it only worked on firefox

Yes, the color changing works in both Firefox and Chrome.

PR #449, opened 2023-09-20
Asterisk spacer

jeryj:
**Description**
Wrapping the spacer in a group block allows the asterisk to be colored via text group, aligned via the justification options, and has no semantic meaning so it is not announced by screen readers.

**Screenshots**
https://github.com/WordPress/twentytwentyfour/assets/967608/2c215560-036b-4757-8834-416776dc1dfd

**Testing Instructions**

1. Add a group block
2. Add a spacer to the group block
3. Select Asterisk Style on the spacer style
4. Change the color with the group block Text Color
5. Change the alignment with the group block Justification
6. Change the size of the asterisk with the group block Text sizes

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -128,6 +128,13 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 				'inline_css' => 'ul.is-style-checkmark-list{list-style-type:"\2713";}ul.is-style-checkmark-list li{padding-inline-start:1ch;}',
 			)
 		);
+		register_block_style(
+			'core/spacer',
+			array(
+				'name'  => 'asterisk',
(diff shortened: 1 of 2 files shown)


scruffian:
@MaggieCabrera @luminuu why was https://github.com/WordPress/twentytwentyfour/pull/471 chosen over this solution?

MaggieCabrera:
> @MaggieCabrera @luminuu why was #471 chosen over this solution?

Because it made more sense semantically that the asterisk is related to the heading. Also, it can be confusing for the users to have the asterisk on the spacer, it's not what you are expecting from this block. Besides the conversations spread all over the 3-4 PRs with alternatives for this I also had a conversation with @richtabor and he agreed that this was blurring the lines of what the block is for

PR #452, opened 2023-09-20
Replace body system font with Inter font

richtabor:
### Description

As mentioned throughout a few issues/prs, let's replace the system font with a more modern (and consistent) font, Inter. The font is also a bit more readable. 

### Before 
CleanShot 2023-09-20 at 17 43 54

### After 
CleanShot 2023-09-20 at 17 44 00


theme.json
```diff
@@ -219,8 +219,19 @@
 			"fluid": true,
 			"fontFamilies": [
 				{
-					"fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
-					"name": "System Font",
+					"fontFace": [
+						{
+							"fontFamily": "Inter",
+							"fontStretch": "normal",
+							"fontStyle": "normal",
```
MaggieCabrera, line 226: should we add an italics version of the font?

MaggieCabrera:
When I open the font file on mac I see that it has all the weights in it. This is something I've seen happen with google fonts, where you select some weights but the download has them all:

Screenshot 2023-09-21 at 09 38 08

Bea suggested to download from [here](https://github.com/rsms/inter/releases/tag/v3.19) instead

luminuu:
Is there a huge difference if we add the italic variation? Or can we just add it as is and maybe add it later if we realise we need it? @richtabor

MaggieCabrera (approved):
I'm bringing this in as it is, we can tweak the font later, but I'd rather have this in the first build and the other Style variations are depending on this one

richtabor:
Yea we can add the italic as well. This one was a variable font; the same pulled from TT3.

PR #456, opened 2023-09-21
Audit template types for insertion options

### LINKED ISSUE #453: Audit template types assigned to each "template" pattern
The patterns prefixed with `template-` are unique in that they are suggested as full templates for replacing/using within the site editor. Let's make sure we're assigning the correct template types for each. 

- Archive alternative templates should only be available on any archive views (including category, tag, author)
- Page alternative templates should only be available for pages, not home/front-page. 
- Search alternative templates, only on search.
- Single templates should only be on page and posts.

We're using `page` a bit too much currently. For example, you wouldn't want to use a "search" template when you've decided to make a "page" template. Note that we do have a mechanism that affords you to choose the kind of template you want to build, so let's make sure the provided results fit the correct decisions. 

CleanShot 2023-09-20 at 18 07 56

jordesign:
Fixes #453 - auditing and fixing the page/post types for which patterns will be suggested as starting points.

I've made a first pass at this. Main remaining question I have is whether we should remove `index` from home-writer and home-portfolio. Seems like we wouldn't want those patterns for an index template- but open to feedback.

### DIFF
--- a/patterns/template-archive-writer.php
+++ b/patterns/template-archive-writer.php
@@ -2,7 +2,7 @@
 /**
  * Title: Writer Archive Template
  * Slug: twentytwentyfour/archive-writer
- * Template Types: archive, page
+ * Template Types: archive, category, tag, author, date
  * Viewport width: 1400
  * Inserter: no
  */
--- a/patterns/template-home-portfolio.php
+++ b/patterns/template-home-portfolio.php
@@ -2,7 +2,7 @@
 /**
  * Title: Portfolio Template
  * Slug: twentytwentyfour/template-home-portfolio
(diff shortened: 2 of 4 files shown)


MaggieCabrera:
Let's hold off on merging this, since I'm not clear if this will render these patterns invisible. Can we test the flows that will let the users actually use these patterns?

MaggieCabrera (approved):

MaggieCabrera:
After reviewing the flows, removing page doesn't make a difference since we have Inserter no, so this makes sense, since page would only apply to the template for all the pages, not just one specific one

PR #457, opened 2023-09-21
Add portfolio template patterns

### LINKED ISSUE #425: Create Index Portfolio (pattern)
The index pattern for the Portfolio variant needs to be created: https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=615-11023&mode=design&t=BUvA3glRfJCamlr0-4

Some work for the archive template has already been done here, just for reference: https://github.com/WordPress/twentytwentyfour/pull/368

### LINKED ISSUE #426: Create Search Portfolio (pattern)
The search pattern for the Portfolio variant needs to be created: https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=615-11262&mode=design&t=BUvA3glRfJCamlr0-4

carolinan:
**Description**
This PR adds:

1. A query loop pattern where the post template only has the featured image block
2. An index template pattern
3. A search template pattern
4. An archive template pattern

**The PR does not attempt to solve the blank spaces for posts that do not have featured images.**

Closes: https://github.com/WordPress/twentytwentyfour/issues/426, 
https://github.com/WordPress/twentytwentyfour/issues/425,
https://github.com/WordPress/twentytwentyfour/issues/351

Repalces https://github.com/WordPress/twentytwentyfour/pull/368

**Testing Instructions**
Search: Open the search template, delete all posts, save and refresh the editor. Select the pattern that only has featured images.
Archive: Open the archive template  delete all posts, save and refresh the editor. Select the pattern that only has featured images.

Index: I was not able to trigger the pattern selection modal only by deleting all the posts.
Instead I had to remove "inserter:no" and insert the portfolio index pattern from the inserter.

### DIFF
--- a/patterns/portfolio-posts.php
+++ b/patterns/portfolio-posts.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Title: List of posts with featured images in three columns
+ * Description: This pattern only shows featured images. Make sure that you have assigned images to your posts.
+ * Slug: twentytwentyfour/portfolio-posts
+ * Categories: query
+ * Block Types: core/query
+ */
+?>
+
+<!-- wp:query {"query":{"perPage":10,"pages":0,"offset":"0","postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true},"align":"wide","layout":{"type":"default"}} -->
+<div class="wp-block-query alignwide">
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50","left":"0","right":"0"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"default"}} -->
+	<div class="wp-block-group" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--50);padding-right:0;padding-bottom:var(--wp--preset--spacing--50);padding-left:0">
+
+		<!-- wp:post-template {"align":"full","style":{"spacing":{"blockGap":"var:preset|spacing|30"}},"layout":{"type":"grid","columnCount":3}} -->
+			<!-- wp:post-featured-image {"isLink":true,"aspectRatio":"3/4","style":{"spacing":{"margin":{"bottom":"0"},"padding":{"bottom":"var:preset|spacing|20"}}}} /-->
+		<!-- /wp:post-template -->
+
+		<!-- wp:spacer {"height":"var:preset|spacing|40","style":{"spacing":{"margin":{"top":"0","bottom":"0"}}}} -->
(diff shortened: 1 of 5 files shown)


carolinan:
> I was not able to trigger the pattern selection modal only by deleting all the posts.

This is also true for the index-writer template.

patterns/template-search-portfolio.php
```diff
@@ -0,0 +1,44 @@
...
+
+<!-- wp:group {"tagName":"main","align":"full","layout":{"type":"constrained"}} -->
+<main class="wp-block-group alignfull">
+	<!-- wp:query-title {"type":"search","align":"wide","style":{"typography":{"lineHeight":"1"},"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|10"}}}} /-->
+
+	<!-- wp:columns {"align":"wide"} -->
+	<div class="wp-block-columns alignwide">
+		<!-- wp:column {"width":"66.66%"} -->
+		<div class="wp-block-column" style="flex-basis:66.66%">
+			<!-- wp:search {"label":"Search","showLabel":false,"width":null,"widthUnit":"%","buttonText":"Search","fontSize":"medium"} /-->
```
MaggieCabrera: I wonder if we should have a hidden pattern for this for translation sake

luminuu: Yeah that would be nice.

carolinan: yeah this whole column can be moved to a pattern and reused.

carolinan: Ah, the column is not used in search-writer, or 404, so, just the single block then.

luminuu: Just merged your other PR with the pattern, you can update this one now

patterns/portfolio-posts.php
```diff
@@ -0,0 +1,36 @@
+<?php
+/**
+ * Title: List of posts with featured images in three columns
+ * Description: This pattern only shows featured images. Make sure that you have assigned images to your posts.
+ * Slug: twentytwentyfour/portfolio-posts
```
MaggieCabrera: I don't care much about the slug but let's rename this file to posts-three-columns so it matches the other 2 query patterns namings

MaggieCabrera:
This is generally looking good but I have one design question. Should the query pattern be wide aligned instead of full width so it aligns with the header and footer? Same goes for the default archive page. @beafialho what do you think?

MaggieCabrera:
> This is generally looking good but I have one design question. Should the query pattern be wide aligned instead of full width so it aligns with the header and footer? Same goes for the default archive page. @beafialho what do you think?

I'm sorry, this is not related to this PR I was in the rust variation that has a different wide width value and it made everything look almost full width. This is looking good! I'll make the naming changes and merge

MaggieCabrera (approved):

PR #458, opened 2023-09-21
Move the link style CSS to theme.json

carolinan:
**Description**
By moving the last remaining CSS to theme.json, the style.css file no longer needs to be enqueued.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -8,32 +8,6 @@
  * @since Twenty Twenty Four 1.0
  */
 
-if ( ! function_exists( 'twentytwentyfour_styles' ) ) :
-
-	/**
-	 * Enqueue styles.
-	 *
-	 * @return void
-	 * @since Twenty Twenty Four 1.0
-	 *
-	 */
-	function twentytwentyfour_styles() {
-		// Register theme stylesheet.
-		wp_register_style(
-			'twentytwentyfour-style',
-			get_template_directory_uri() . '/style.css',
-			array(),
-			wp_get_theme()->get( 'Version' )
-		);
-
-		// Enqueue theme stylesheet.
-		wp_enqueue_style( 'twentytwentyfour-style' );
-	}
-
-endif;
-
-add_action( 'wp_enqueue_scripts', 'twentytwentyfour_styles' );
-
 /**
  * Register block Styles
  */
--- a/functions.php
+++ b/functions.php
@@ -8,25 +8,6 @@
  * @since Twenty Twenty Four 1.0
  */
 
-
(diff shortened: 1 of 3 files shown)


MaggieCabrera:
We've completely forgotten RTL here, we might need to get back to this when we do, maybe even enqueue another file... I'm not too happy about that

carolinan:
Yes maybe. Lets add the don't merge label
-I would prefer if a native speaker tested the RTL of the theme though.

MaggieCabrera:
> Yes maybe. Lets add the don't merge label -I would prefer if a native speaker tested the RTL of the theme though.

That's a great idea, where should we ask for help? polyglots?

carolinan:
Yes

aristath (approved):
IMO this is a great change and promotes the new paradigm of how a block theme _should_ be done.

The code looks good and works as expected.

Regarding RTL, I don't see anything in the CSS here that would require any RTL feedback... Gutenberg is _supposed to_ handle RTL styles etc on its own. If that is not the case, then that's something that should probably be fixed in Core/Gutenberg and not the theme itself since it would happen on all themes and not just this one.

The theme doesn't (currently) add any CSS for left/right paddings/margins/borders etc, so in theory it doesn't need any special tweaks for RTL.

I agree 100% that it should be tested in RTL, but that should not be considered a blocker for this PR since it doesn't touch anything related to it 👍

MaggieCabrera:
> IMO this is a great change and promotes the new paradigm of how a block theme _should_ be done.
> 
> The code looks good and works as expected.
> 
> Regarding RTL, I don't see anything in the CSS here that would require any RTL feedback... Gutenberg is _supposed to_ handle RTL styles etc on its own. If that is not the case, then that's something that should probably be fixed in Core/Gutenberg and not the theme itself since it would happen on all themes and not just this one.
> 
> The theme doesn't (currently) add any CSS for left/right paddings/margins/borders etc, so in theory it doesn't need any special tweaks for RTL.
> 
> I agree 100% that it should be tested in RTL, but that should not be considered a blocker for this PR since it doesn't touch anything related to it 👍

I agree that this is what I would want to do too, but we do have things that may affect RTL, like the checkmarks block style. Now that I'm looking at it I wanted to test it and that has stopped working, seems like the CSS for it is not loading after we moved the CSS to be inline.

Looking at the CSS, I think you might be right and we won't need an RTL stylesheet after all. I was recalling an earlier version of that CSS that positioned the checkmarks differently.

I'll get this one in and we can always revisit if we feel we need to. I've asked for an RTL review anyway just in case, and we should look into why the checkmarks stopped working on a different PR.

PR #459, opened 2023-09-21
Replace some images with placeholders

carolinan:
**Description**
Closes https://github.com/WordPress/twentytwentyfour/issues/392

**Screenshots**

**Testing Instructions**
Place the updated patterns in the block editor and confirm that there are no issues. Save and view the front. The images should not show.

Issue #364, opened 2023-09-13
Ensure all assets have proper credits/licensing information


sabernhardt:
Ideally before merging into Core:

1. Check the sources of all bundled fonts and images.
2. Add credits to the readme.txt.
3. Add any necessary license.txt files.

sabernhardt:
Right now, none of the images have credits in the readme.

The Cardo font has credit and licensing information, with a link to the license. Should it have a copy of the license document too?

MaggieCabrera:
Thank you for logging this issue, it's part of the pre launch tasks. I checked TT3 and it's [including the license](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-content/themes/twentytwentythree/assets/fonts/dm-sans/LICENSE.txt) files inside the assets folders, so I'm happy to go with that.

TT2 does [this](https://github.com/WordPress/wordpress-develop/blob/11c53d559bf89c4647f75ca464886d592c0c1fb1/src/wp-content/themes/twentytwentytwo/readme.txt#L89) about the images. There's still some potential changes coming to the images if we remove some to use placeholders instead, but it's unlikely that we will be adding more, so this is a task we could start working on. @beafialho, can you tell us where to find the info for the image licensing? If you download a zip with the theme you can check the images being used inside the assets folder, we only need info for those

beafialho:
Here are the image sources for the images in the assets folder:

**about.webp** - https://www.rawpixel.com/image/3297419/free-photo-image-interior-hallway-architecture
**business-app.webp** - https://www.rawpixel.com/image/5928004/photo-image-public-domain-hand-person
**business-hero.webp** - https://www.rawpixel.com/image/430335/free-photo-image-architecture-building-roof
**home.webp** - https://www.rawpixel.com/image/3286615/free-photo-image-texture-architecture-building
**left-aligned-cta.webp** - https://www.rawpixel.com/image/3283970/free-photo-image-pattern-artwork-abstract
**project-1.webp** - https://www.rawpixel.com/image/3289063/free-photo-image-architectural-detail-architecture
**project-2.webp** - https://www.rawpixel.com/image/430585/free-photo-image-stairs-staircase-concrete
**project-overview.webp** - https://www.rawpixel.com/image/3286187/free-photo-image-person-stairs-architecture
**rsvp.webp** - https://www.rawpixel.com/image/430558/free-photo-image-abstract-architecture-art-public-domain
[...]

PR #460, opened 2023-09-21
Add image license

carolinan:
**Description**
Adds image license and image source links to the readme.txt file.

Partial for https://github.com/WordPress/twentytwentyfour/issues/364
I did not add the images that are removed in https://github.com/WordPress/twentytwentyfour/pull/459

@beafialho What is the license and source of icon-message.webp ?


kafleg:
I think we also need to add, https://www.rawpixel.com/services/licenses 
This link is for raw pixel image license.

carolinan:
I think that may be confusing since it includes more than one licence type and the theme is only using one'?

beafialho:
@carolinan the `icon-message.webp` is a unicode symbol which was exported as an image.

carolinan:
I have never tried to look up licensing for unicode.. is it this? https://www.unicode.org/license.txt From https://www.unicode.org/policies/licensing_policy.html

beafialho:
> I have never tried to look up licensing for unicode

Me neither, but this seems to be it yes.

luminuu:
@carolinan Can this PR be merged or do you want to add the remaining image licenses from stocksnap.io here too?

carolinan:
@luminuu It can be merged, there are two PR's to remove those images

luminuu (approved):
LGTM

PR #463, opened 2023-09-21
Fix Portfolio Post Template: Misplaced post meta

carolinan:
**Description**
I moved the post meta template part from the right to the left column, and added the theme slug to the template parts.
I replaced the footer with the default footer.

**Screenshots**
Single portfolio

**Testing Instructions**
In the Site Editor, select the single post template. Delete all blocks. Save. Refresh to trigger the pattern selection modal. Add the correct pattern and view the post meta.

### DIFF
--- a/patterns/template-single-portfolio.php
+++ b/patterns/template-single-portfolio.php
@@ -23,11 +23,11 @@
 
 	<!-- wp:columns {"align":"wide"} -->
 	<div class="wp-block-columns alignwide"><!-- wp:column {"width":"45%"} -->
-	<div class="wp-block-column" style="flex-basis:45%"></div>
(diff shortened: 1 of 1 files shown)


patterns/template-single-portfolio.php
```diff
@@ -23,11 +23,11 @@
 
 	<!-- wp:columns {"align":"wide"} -->
 	<div class="wp-block-columns alignwide"><!-- wp:column {"width":"45%"} -->
-	<div class="wp-block-column" style="flex-basis:45%"></div>
+	<div class="wp-block-column" style="flex-basis:45%"><!-- wp:template-part {"slug":"post-meta","theme":"twentytwentyfour"} /--></div>
```
MaggieCabrera: I think we don't need columns anymore, can we just make this a group?

MaggieCabrera (approved):

PR #464, opened 2023-09-21
Heading on left: Fix missing closing tag

carolinan:
**Description**
Fix a missing p tag that was causing a block validation error

 props @kafleg


carolinan:
@MaggieCabrera I don't like the name of this pattern, since its a paragraph not a heading....

MaggieCabrera:
> @MaggieCabrera I don't like the name of this pattern, since its a paragraph not a heading....

I agree, I think we copied from the figma names. Feel free to open a PR with a suggestion. there's probably other patterns that don't have the best names either.

aristath (approved):

Issue #420, opened 2023-09-19
Style Variation: Sandstorm


beafialho:
<img width="873" alt="Captura de ecrã 2023-09-19, às 11 59 39" src="https://github.com/WordPress/twentytwentyfour/assets/65220155/7953d662-80a9-407f-acea-9222653f88f7">
<img width="872" alt="Captura de ecrã 2023-09-19, às 11 59 32" src="https://github.com/WordPress/twentytwentyfour/assets/65220155/da608834-1391-490d-bea6-1cd58494e02a">

[sandstorm.json.zip](https://github.com/WordPress/twentytwentyfour/files/12659019/sandstorm.json.zip)

MaggieCabrera:
We have a problem with the fonts in that if we have style variations that use the same font for heading and body they will show duplicated on the font families dropdown :( We decided to go around that by using Inter for the headings and the system font for the body on the sans-serif variations that you designed (discussed in https://github.com/WordPress/twentytwentyfour/pull/367#issuecomment-1721272112)

beafialho:
Oh, that's weird... This variation uses **Instrument Sans** though.

MaggieCabrera:
> Oh, that's weird... This variation uses **Instrument Sans** though.

I know, but the result of having it apply to both body and headings will end up with two instances of Instrument sans, one for each of them

beafialho:
So, we can't have a variation with just one font? Why is it a problem that it shows duplicated?

MaggieCabrera:
> So, we can't have a variation with just one font? Why is it a problem that it shows duplicated?

just that it's confusing for the user, and also because we are enqueing the font twice, which is not great but also not terrible. @richtabor what do you think?

richtabor:
Yea it’s a strange edge case we’re hitting. Perhaps we should open an issue to never duplicate fonts. I don’t expect it for 6.4, but would be good anyhow. 

Could we use inter for body font here as well?

beafialho:
Let me rethink the variations. I'd love to use some different fonts.

beafialho:
Ok I think I found a nice solution for this one, using Instrument Sans for paragraphs and Jost for headings. Updated file here:

[sandstorm-2.json.zip](https://github.com/WordPress/twentytwentyfour/files/12674152/sandstorm-2.json.zip)

[...]

PR #468, opened 2023-09-21
Add Mint

### LINKED ISSUE #419: Style Variation: Mint
Captura de ecrã 2023-09-19, às 11 58 22
Captura de ecrã 2023-09-19, às 11 58 15

[mint.json.zip](https://github.com/WordPress/twentytwentyfour/files/12659011/mint.json.zip)

carolinan:
**Description**
Copied from Mint 2: https://github.com/WordPress/twentytwentyfour/issues/419


luminuu:
I think the fonts are missing as they're referred to Instrument Sans and Jost, @beafialho and @MaggieCabrera have been discussing them in #420

luminuu:
Added the fonts Instrument Sans and Jost to the main `theme.json` file and updated the style variation file a bit.

styles/mint.json
```diff
@@ -0,0 +1,115 @@
...
+				"typography": {
+					"fontFamily": "var(--wp--preset--font-family--instrument-sans)",
+					"fontSize": "var(--wp--preset--font-size--large)",
+					"fontStyle": "normal",
+					"fontWeight": "400"
+				}
+			},
+			"core/site-title": {
+				"typography": {
+					"fontFamily": "var(--wp--preset--font-family--instrument-sans)"
```
MaggieCabrera: if we do this this way it won't work with the rest of the patterns that are using heading/body naming for their fonts. Why are we defining the font on the global theme.json if it's not using it? We should do it on the variation instead. Core will be adding a way for users to add new fonts soon so the theme doesn't need to worry about providing the user with extras

luminuu: I've compared it with TT3 where all fonts were put in the theme.json. Haven't thought about the heading/body naming. Just added an update.

MaggieCabrera (approved):
I made the changes to the fonts so they fit the design and work with the patterns

PR #469, opened 2023-09-21
Translations and post meta for Three Featured posts pattern

MaggieCabrera:
**Description**

This pattern was using the old markup for the post meta so I added the template part and included translation functions for the heading


luminuu (approved):
LGTM

PR #470, opened 2023-09-21
Fix checkmarks block style

### LINKED ISSUE #466: Checkmarks block style for list block stopped working
At some point this block style stopped working. The class is present and you can still apply it in the editor but the CSS is not being applied to it:

Screenshot 2023-09-21 at 10 34 57

To test this insert a list block and select the Checkmark variation. The discs should turn into checkmarks. The rest of the variations seem to be working fine, including the one with the arrow for the details block, which has a similar implementation.

MaggieCabrera:
**Description**

Fixes https://github.com/WordPress/twentytwentyfour/issues/466

For some reason, this block style had the wrong arg name for the inline CSS. This should fix it

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -74,8 +74,8 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 		register_block_style(
 			'core/list',
 			array(
-				'name'       => 'checkmark-list',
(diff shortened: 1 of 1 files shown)


luminuu (approved):
Tested and works 👍

PR #471, opened 2023-09-21
Asterisk as heading variation

MaggieCabrera:
**Description**

Another idea for https://github.com/WordPress/twentytwentyfour/pull/51

Closes #51

Following @carolinan input that this makes more sense semantically to belong to the heading if we are going to follow the block style route, which I agree. The user won't have control over the color, it will always use Contrast 2, which is NOT the same color as the heading, as per the design. There's no way around this. Also we need to have RTL in mind, I have added CSS for centered and right alignment, but we need to see what happens for RTL.

So far this is my preferred option of all the ones we've explored, since it covers most of the requirements and feels a little less hacky or confusing for users. And it doesn't add any markup for screen readers.

**Screenshots**

Screenshot 2023-09-21 at 12 57 19

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -86,8 +86,8 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 		register_block_style(
 			'core/heading',
 			array(
-				'name'       => 'asterisk',
-				'label'      => __( 'With asterisk', 'twentytwentyfour' ),
+				'name'         => 'asterisk',
+				'label'        => __( 'With asterisk', 'twentytwentyfour' ),
 				/* TODO: RTL considerations */
 				'inline_style' => ".is-style-asterisk:before{content:'';width: 1.5rem;height:3rem;background: var(--wp--preset--color--contrast-2);
 					clip-path: path('M11.93.684v8.039l5.633-5.633 1.216 1.23-5.66 5.66h8.04v1.737H13.2l5.701 5.701-1.23 1.23-5.742-5.742V21h-1.737v-8.094l-5.77 5.77-1.23-1.217 5.743-5.742H.842V9.98h8.162l-5.701-5.7 1.23-1.231 5.66 5.66V.684h1.737Z'); display: block;}.is-style-asterisk.has-text-align-center:before{margin: 0 auto;}.is-style-asterisk.has-text-align-right:before{margin-left: auto;}.rtl .is-style-asterisk.has-text-align-left:before{margin-right: auto;}",
--- a/functions.php
+++ b/functions.php
@@ -83,6 +83,16 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 				'inline_css' => 'ul.is-style-checkmark-list{list-style-type:"\2713";}ul.is-style-checkmark-list li{padding-inline-start:1ch;}',
 			)
 		);
+		register_block_style(
+			'core/heading',
+			array(
+				'name'       => 'asterisk',
+				'label'      => __( 'With asterisk', 'twentytwentyfour' ),
+				/* TODO: RTL considerations */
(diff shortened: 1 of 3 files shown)


MaggieCabrera:
As this is a proof of concept, I didn't delete the older glyphs and I didn't apply it to the other pattern that uses this. I will do so if we get positive feedback on the approach

luminuu:
I just tested this and like this approach the most.

carolinan:
On the front, the body has a class called `rtl` which can be used, but this does not work in the iframed editor.

carolinan:
RTL:
Asterisk is on the right, editor
Front:
Asterisk is on the right, front

MaggieCabrera:
ooh, that looks good, what happens if you align the heading to the right in Arabic manually?

carolinan:
It is right aligned by default.
https://github.com/WordPress/twentytwentyfour/assets/7422055/8b36d5d0-d65a-4a9f-83ff-8ed333edef25

MaggieCabrera:
> It is right aligned by default. https://github.com/WordPress/twentytwentyfour/assets/7422055/8b36d5d0-d65a-4a9f-83ff-8ed333edef25

I suspected that would happen. I will fix it for the front then and hopefully when we have classes in the iframe it will be fixed for the editor. It's kind of an edge case that is not apparent unless you play with the block yourself, so I think it's acceptable if the frontend shows fine

carolinan:
Would it not be enough to always support both left and right alignment?

luminuu:
I just tested this in a RTL language and it's also not working in the frontend with the left alignment:

CleanShot 2023-09-21 at 13 37 52@2x

MaggieCabrera:
> Would it not be enough to always support both left and right alignment?

I don't know how we could do that to work both on RTL and LTR languages, since I'm using margins to position it. If you think we can position the asterisk some other way, please let me know, I couldn't come up with a way that would simply inherit the text-align rule while keeping the asterisk above the text.

MaggieCabrera:
I added the rtl fix for the frontend:

LTR
Screenshot 2023-09-21 at 15 39 30

RTL
Screenshot 2023-09-21 at 15 40 45

MaggieCabrera:
I extended the changes to the other pattern in case we want to bring this in. Now I'm feeling protective of these little asterisks 😆

luminuu:
Perfect, that looks good. Do we want to wait for further feedback or do you think we should just merge it?

MaggieCabrera:
If you two like it, I like it, then I say go ahead. We can refine the CSS if we find a cleaner way for the positioning on a follow up, there might be some smarter way to go about it

MaggieCabrera:
I see the rust variation is not using a contrast 2 color, I'm trying to provide a fallback to base, but it doesn't look like `background: var(--wp--preset--color--contrast-2, var(--wp--preset--color--base, transparent));` is working

MaggieCabrera:
I pushed a fix to add a fallback for variations that don't define a contrast 2 color, this will default to use `currentColor`. Rust is an example of that case

Screenshot 2023-09-21 at 16 37 36

jeryj:
This is working well in my testing! The fallback to currentColor is nice too.

One concern I have, which may not be too serious, is that if people want to display an asterisk without a heading, they'll use an empty heading block with asterisk styles. This gets them the asterisk without the visual heading, and they won't realize this will announce the heading element even though it's empty. I'm not sure how common it is for people to use theme designs in unintended ways though.

luminuu:
That's a good point, however I don't think we can find a solution that is completely accessibility ready in every use case. It is definitely better than the other options. I'd say it's a rare use case but we can point to this issue in the documentation.

MaggieCabrera:
this still needs a green check, I'd like to bring this one in

luminuu (approved):

PR #472, opened 2023-09-21
Add a hidden search pattern

carolinan:
**Description**
Replaces the search block with a pattern with translatable label and button text.

**Testing Instructions**
View each of the updated patterns and confirm that the search block still works correctly:
404,
search,
search-writer


luminuu (approved):
LGTM

Issue #356, opened 2023-09-13
RTL support and review


MaggieCabrera:
Let's make time to review the whole theme using a RTL language to see that things are placed on the correct places. Particularly things like the markers and icons:

<img width="1025" alt="Screenshot 2023-09-13 at 17 47 32" src="https://github.com/WordPress/twentytwentyfour/assets/3593343/dca28939-cb9e-401d-928e-99765ec128b0">

vraja-pro:
Tested list and details blocks with RTL (Hebrew), no issue detected.
![Screenshot 2023-10-04 at 14 43 11](https://github.com/WordPress/twentytwentyfour/assets/65466507/f57752b4-052c-4b11-bd91-cf2726b987fb)
@carolinan

MaggieCabrera:
@vraja-pro thank you for testing!

MaggieCabrera:
I think we can close this one, let's reopen if we find any issues!

PR #481, opened 2023-09-21
Remove right margin from Pill style links

sabernhardt:
**Description**

In #320, the Pill-style taxonomy links were given both bottom and right margins.

The right margin is not mirrored for right-to-left languages (#356), but I do not think the side margin is necessary.

If more space is desired, maybe giving the `.wp-block-post-terms__separator` element a `min-width` could be better.

**Screenshots**

Tags blocks with right margin (LTR)

Tags with right margin in English

Tags blocks, removing the side margin (LTR)

Tags without the side margin in English

Tags blocks with right margin (RTL)

Tags with right margin, setting the language to Hebrew

Tags blocks, removing the side margin (RTL)

Tags without the side margin, using Hebrew as the RTL language

**Testing Instructions**

1. Activate the theme.
2. Create a post.
3. Insert a Tags block (or another taxonomy).
4. Add tags to the post.
5. Save the post and view it on the front.
6. Go to the Site Editor and select the Single Posts template. If you do not already have a Tags block in that template, add one.
7. Visit the post again.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -68,7 +68,7 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 				 * Styles variation for post terms
 				 * https://github.com/WordPress/gutenberg/issues/24956
 				 */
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
I don't think we need margin at all, there are already separators there. I added some other tweaks since we were using hardcoded colors for this and it didn't work on the style variations of the theme.

MaggieCabrera (approved):

PR #483, opened 2023-09-21
Underline post meta

luminuu:
Based on the discussions in #27, this is a suggestion to underline the post meta links, as they can be considered content and should be underlined for accessibility reasons.

CleanShot 2023-09-21 at 20 16 49@2x

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -9,42 +9,27 @@
 			"defaultGradients": false,
 			"duotone": [
 				{
-					"colors": [
-						"#111111",
-						"#ffffff"
-					],
+					"colors": ["#111111", "#ffffff"],
 					"slug": "duotone-1",
 					"name": "Black and white"
 				},
 				{
-					"colors": [
-						"#111111",
-						"#C2A990"
-					],
+					"colors": ["#111111", "#C2A990"],
 					"slug": "duotone-2",
 					"name": "Black and sandstone"
 				},
 				{
-					"colors": [
-						"#111111",
-						"#D8613C"
-					],
+					"colors": ["#111111", "#D8613C"],
 					"slug": "duotone-3",
 					"name": "Black and rust"
 				},
 				{
-					"colors": [
-						"#111111",
-						"#B1C5A4"
-					],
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
From that same original issue:

> Links must be underlined when they're embedded within other text, such as within the body content of an article or comment.

>  Areas such as post meta data can fall into a broad spectrum of importance, depending on other clarifying indicators. I think we, as WordPress developers, tend to automatically identify post meta items such as author and category links as links, because we know internally how these are structured and what kind of content is there. Whether the design supports that is much more subjective.

I don't know if this is the case. We should be underlining the post titles too if that were the case. Looking at TT3, it only underlines the actual categories and tags. Looking at TT2 it underlines everything. I'd like a design eye on this too.

luminuu:
Ping @beafialho @richtabor any thoughts on this?

beafialho:
The design was thought out not to have post meta underlined, please let's leave this as it is. 

It's an exception, as mentioned [here](https://www.nngroup.com/articles/guidelines-for-visualizing-links/): _"Assuming the link text is colored, it's not always absolutely necessary to underline it"._

jasmussen:
These links are not embedded within other text. They are on a line of their own, feature a smaller font size, and are separated by em-dashes. 

The lack of an underline on meta items is an intentional design choice that is meant to reduce the prominence of those links, as they will intrinsically be less important than links in the body text. This design choice of balancing prominence is ultimately about creating [simple and consistent layouts](https://accessibility.blog.gov.uk/2016/09/02/dos-and-donts-on-designing-for-accessibility/), which is inclusive of users that need clarity.

I would recommend not making this change.

beafialho:
Also, notice you have a dot after the day:

Captura de ecrã 2023-09-22, às 11 18 51

luminuu:
> Also, notice you have a dot after the day:
> Captura de ecrã 2023-09-22, às 11 18 51

That is because the date setting is `j. F Y` in WP (standard German date format)

Thanks everyone for the input, I'm closing this PR.

PR #486, opened 2023-09-22
Make the default home page a pattern so it can be switched back to

scruffian:
**Description**

If we land https://github.com/WordPress/gutenberg/pull/54609, we need a way for users to preview the original home template.

**Screenshots**

Screenshot 2023-09-22 at 10 38 42

**Testing Instructions**

1. Checkout https://github.com/WordPress/gutenberg/pull/54609
2. Open the "Replace template" option
3. Confirm that you see the default template as well as the writer and portfolio ones.

**Contributors**

### DIFF
--- a/patterns/template-home-business.php
+++ b/patterns/template-home-business.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Home Template
+ * Title: Business Home Template
  * Slug: twentytwentyfour/template-home
  * Template Types: front-page, index, home
  * Viewport width: 1400
--- a/patterns/template-home-portfolio.php
+++ b/patterns/template-home-portfolio.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Portfolio Template
+ * Title: Portfolio Home Template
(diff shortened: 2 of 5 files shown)


MaggieCabrera:
I think you are missing files here :D

scruffian:
you're right, added!

carolinan:
We should do this regardless of whether the Gutenberg PR will be ready or not.
I am not convinced it is a good idea to include the feature in 6.4 since it is not complete and won't be tested in time, and I don't see a "blessing" from the release leads.

luminuu:
We can try to escalate this to the release leads if it makes sense to do so.

MaggieCabrera:
> We can try to escalate this to the release leads if it makes sense to do so.

Yeah, since this release is so short, I think we can push for this to actually make it. If testing during beta reveals it's not ready, we can always pull it back

luminuu:
Just noticed we have the linked issue to this PR already in our Gutenberg tracking issue, which has been blessed last week. So I hope we can get it in!

MaggieCabrera (approved):
I renamed the patterns for consistency, LGTM

PR #487, opened 2023-09-22
Modifying the date format at hidden-comments.php

emdashftw:
**Description**
The date format of a comment includes a hardcoded "at" word that's used in other locales.

That's a problem while using other locales ie. in Spanish, where the `at` word doesn't exist and means nothing.

Modifying it with "-" makes it generic for all locales.

I believe that using a `-` instead of `at` would be a saner default. Since `-` is generic and doesn't depend on the current language being used.

**Screenshots**
Before:
image

After:
image

### DIFF
--- a/patterns/hidden-comments.php
+++ b/patterns/hidden-comments.php
@@ -22,7 +22,7 @@
 				<!-- wp:group -->
 				<div class="wp-block-group">
 					<!-- wp:comment-author-name /-->
-					<!-- wp:comment-date {"format":"F j, Y - g:i a"} /-->
+					<!-- wp:comment-date /-->
 				</div>
 				<!-- /wp:group -->
 			</div>
--- a/patterns/hidden-comments.php
+++ b/patterns/hidden-comments.php
@@ -22,7 +22,7 @@
 				<!-- wp:group -->
 				<div class="wp-block-group">
 					<!-- wp:comment-author-name /-->
-					<!-- wp:comment-date {"format":"F j, Y \\a\\t g:i a"} /-->
+					<!-- wp:comment-date {"format":"F j, Y - g:i a"} /-->
 				</div>
 				<!-- /wp:group -->
 			</div>


patterns/hidden-comments.php
```diff
@@ -22,7 +22,7 @@
 				<!-- wp:group -->
 				<div class="wp-block-group">
 					<!-- wp:comment-author-name /-->
-					<!-- wp:comment-date {"format":"F j, Y \\a\\t g:i a"} /-->
+					<!-- wp:comment-date {"format":"F j, Y - g:i a"} /-->
```
MaggieCabrera: we could also wrap the whole thing in a translation function and let each translator decide?

emdashftw: I was unsure on what translation function to use since I'm not familiarized with wordpress i18n, so I choosed keeping it simple initially (in my first ever pull request btw).

But I guess it should be as simple as doing:

`<!-- wp:comment-date {"format":"F j, Y <?php echo esc_html__( '\\a\\t', 'twentytwentyfour' ); ?> g:i a"} /-->`

or

`<!-- wp:comment-date {"format":"F j, Y <?php esc_html_e( '\\a\\t', 'twentytwentyfour' ); ?> g:i a"} /-->`

MaggieCabrera: if you check the [README's guidelines](https://github.com/WordPress/twentytwentyfour/#pattern-creation-guidelines) you see the options you have. esc_html_x is probably the best option here since it allows you to add context for the translators (check how other strings have done this over the theme). I would include the whole date format string, because for example, the comma after the day may not be something every language wants to include.

luminuu: Will it be properly rendered then if it's in a translation function?

MaggieCabrera: good question, I don't know what format is expecting, we could look into that, usually themes don't bother adding a custom format at all. Also it might as well be that GB already takes care of translating the format and we shouldn't touch this at all, I'll go check

luminuu: No I wouldn't want to put the entire date string into the translation function, as there are separate settings for date and time format in WordPress, and I'm worried they won't be taken into consideration when you hardcode it into a translation string.

luminuu: I just spotted some more issues with the dates comparing the post patterns on the home page, post meta on single and comment date.

MaggieCabrera: Looks like it was a custom format instead of choosing one of the presets:

Screenshot 2023-09-22 at 12 36 54

To be honest, given that it's the comment form I'd remove the format attr altogether and let the block do the work for us, that's what's been done on previous themes

luminuu: If the format is set to `{"format":""}`, it will fall back to the date and time format settings found in the general settings of WordPress. We should go for this one instead of forcing certain date and time formats in the theme itself, as they vary between countries.

emdashftw: I've read it, and I still was unsure haha.

But I see that in the same file on line 13 was done like this:
`<h2 class="screen-reader-text"><?php esc_html_e( 'Comments', 'twentytwentyfour' ); ?></h2>`

So that's another thing to consider, for consistence.

I tried doing it like in my comment above and it seems to work fine.

Again I have doubts that including the whole format string would make sense as a translation. I mean, a translation should include a format decision ?

About the format itself, what I can add is that in the US they often use mm/dd/yyyy while in europe is more common to use dd/mm/yyyy so yeah, the comma and the whole format... would depend on the culture of the user.

luminuu: Just remove the format (see my comment above), don't add any translation stuff. This should be handled completely by WP itself. If the user wants a custom format, they can go in the editor and change it (see screenshot above by Maggie).

emdashftw: Oh, I see. You both were faster than me commenting. 😓

Would you keep the format property:
`<!-- wp:comment-date {"format":""} /-->`

or we would just do:
` <!-- wp:comment-date /-->`

Completely removing it also falls back to general settings, so I don't see any reason to keep the `{"format":""}` part.

luminuu: If it falls back to the default without the format attribute, I'm fine leaving it out.

emdashftw (commented):
I've submitted a new commit removing the format from comment-date. Since we've talked this can be handled completely by wordpress, and the format can vary through countries / locales.

_As I stated before, is my very first pull request. So excuse me please, this is pretty new to me. If I understood it correctly, one of you should approve the review in order to merge right ? It was correctly requested ? Is there anything else I'm missing ?_

MaggieCabrera (approved):

MaggieCabrera:
> I've submitted a new commit removing the format from comment-date. Since we've talked this can be handled completely by wordpress, and the format can vary through countries / locales.
> 
> _As I stated before, is my very first pull request. So excuse me please, this is pretty new to me. If I understood it correctly, one of you should approve the review in order to merge right ? It was correctly requested ? Is there anything else I'm missing ?_

That's right, we take it from here, thanks a lot for contributing!

Issue #489, opened 2023-09-22
[Pattern] Centered Statement: Improvement


beafialho:
I noticed this pattern wasn't looking like the designs, and the text also had an extra dash which shouldn't be there so I made the adjustments, the markup is below:

| Currently      | Designs |
| ----------- | ----------- |
|  <img width="1283" alt="Captura de ecrã 2023-09-22, às 11 53 00" src="https://github.com/WordPress/twentytwentyfour/assets/65220155/8aaed1e3-b285-403a-8d9d-76591222107b">    |    <img width="1285" alt="Captura de ecrã 2023-09-22, às 11 52 41" src="https://github.com/WordPress/twentytwentyfour/assets/65220155/bca8e3b9-d4e2-4a3d-ae12-6c5bb162ebb9">    |

```
<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60","left":"var:preset|spacing|60","right":"var:preset|spacing|60"},"margin":{"top":"0","bottom":"0"}}},"backgroundColor":"base-2","layout":{"type":"constrained"}} -->
<div class="wp-block-group alignfull has-base-2-background-color has-background" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)"><!-- wp:group {"align":"wide","layout":{"type":"default"}} -->
<div class="wp-block-group alignwide"><!-- wp:spacer -->
<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer -->

<!-- wp:paragraph {"align":"center","style":{"typography":{"lineHeight":"1.2","fontStyle":"normal","fontWeight":"400"}},"fontSize":"x-large","fontFamily":"heading"} -->
<p class="has-text-align-center has-heading-font-family has-x-large-font-size" style="font-style:normal;font-weight:400;line-height:1.2"><em>Études</em> is not confined to the past—we are passionate about the cutting edge designs shaping our world today.</p>
<!-- /wp:paragraph -->

<!-- wp:spacer -->
<div style="height:100px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer --></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```

[...]

PR #496, opened 2023-09-22
Updated Centered Statement Pattern

shail-mehta:
**Fixes:**  https://github.com/WordPress/twentytwentyfour/issues/489


beafialho:
Looks great, thank you!

carolinan:
Is the set pixel height a requirement here? I am asking because there was an effort earlier to replace these values with the presets.

Similarly, there are styles for the paragraphs that  are unnecessary in the markup because they are the default values. These have already been removed from theme at least once.

MaggieCabrera:
> Similarly, there are styles for the paragraphs that are unnecessary in the markup because they are the default values. These have already been removed from theme at least once.

Those are apparently being added by the editor when selecting a color for a paragraph, they appear when copying the markup after doing that, I don´t think Bea is doing anything to the code that the editor is generating.

+1 on the pixels/preset issue, maybe we didn't do a good job selecting the presets that are closer to the design intention?

beafialho:
Can we find a viable solution to have the pattern look like the design?

MaggieCabrera:
I changed the spacer to use the preset 50, which looks almost the same in height as 100px on desktop. I think the rest looks fine and the pattern looks as Bea intended:

Screenshot 2023-09-29 at 11 42 23

@beafialho does this look good to you?

beafialho:
Yes, looks great. Thank you!

MaggieCabrera (approved):

PR #497, opened 2023-09-22
Update TTF to WOFF2 and Rename Font File Name

huzaifaalmesbah:
I update TTF to WOFF2 and rename font file name


luminuu:
Is renaming the font files really needed? What benefit does it have?

MaggieCabrera:
> Is renaming the font files really needed? What benefit does it have?

+1, I think it's fine to leave the name as it was, since it shows it's a variable font

huzaifaalmesbah:
> +1, I think it's fine to leave the name as it was, since it shows it's a variable font

**Then we can use like this.**
InstrumentSans-Variable.woff2
InstrumentSans-Italic-Variable.woff2

luminuu (approved):
Updated the font files to the original names, and fixed the folder for the Jost font, as it was still `instrument-sans`. Fonts are now loading correctly.

PR #499, opened 2023-09-25
Experiment: Featured image fallback

carolinan:
**Description**
This PR filters the `` tag of the post featured image block when the post or page does not have a featured image.
It adds a 1px x 1px white (#FFFFFF) png file as the image source.
It adds the title of the post as the alt attribute.

**Idea:** Try a transparent image.

**Screenshots**

Home, default style variation:
Featured image fallback on the home template

Single post:
Featured image fallback on the single post

Archive:
Featured image fallback on single post

Home, Onyx:
Featured image fallback on the Home of the Onyx style variation

**Portfolio** home template, onyx style.
This time, with more content, using the theme test data import.
portfolio home, onyx

**Portfolio** home template, milky style, mobile (cut off because the image was too large)
portfolio home, milky

**Testing Instructions**
Make sure that your WordPress installation has posts with and without featured images.
View different content that shows featured images:
The home templates, the portfolio templates, single posts with featured images, etc.
Try different browser widths and style variations.

**Contributors**
greenshady justintadlock


carolinan:
### Resulting markup (front)
Copied from the default archive.

Regular featured image block:
`<figure style="aspect-ratio:3/4; padding-bottom:var(--wp--preset--spacing--20);margin-bottom:0;" class="wp-block-post-featured-image"><a href="http://64.local/2023/09/23/has-featured-image/" target="_self">Has featured image</a></figure>`

Filtered:
`<figure style="aspect-ratio:3/4; padding-bottom:var(--wp--preset--spacing--20);margin-bottom:0;" class="wp-block-post-featured-image"><a href="http://64.local/2023/09/23/hello-world/" target="_self">Hello world!</a></figure>`

MaggieCabrera:
I'm not sure it's a good assumption to make that we are expecting everyone to have a featured image on all their posts

carolinan:
I don't understand. We are assuming they are not adding it to every post, that is why the fallback image is used.

MaggieCabrera:
> I don't understand. We are assuming they are not adding it to every post, that is why the fallback image is used.

mmh, what I mean is that if there's no featured image, nothing should show, not an empty white box that looks like something is missing. That feels like something is wrong with the website

carolinan:
Well we tried removing the posts through a query loop variation and it did not work / was not suitable for the archives.

carolinan:
Portfolio archive without this pr:
https://github.com/WordPress/twentytwentyfour/assets/7422055/52306cd4-5368-4a30-aabf-04558e80fa20

With this pr:
https://github.com/WordPress/twentytwentyfour/assets/7422055/d96e65e3-9805-4e83-a177-4195ac8e6574

carolinan:
I tried to see if I could limit it to the portfolio templates, but there is no way to identify what pattern is in the template.

luminuu:
I personally like that idea, it gives a more rounded touch than just having nothing show up at all, which would also mean there's nothing to click on if there's no title, like in the portfolio templates.

Maybe @beafialho can give some design input here?

MaggieCabrera:
> I personally like that idea, it gives a more rounded touch than just having nothing show up at all, which would also mean there's nothing to click on if there's no title, like in the portfolio templates.
> 
> Maybe @beafialho can give some design input here?

I'm not sure. It looks better than an empty image on the portfolio view but the single post I really don't like, it feels like a bug, like something is wrong. The other way you only get that feeling for the archive template, which we provide alternatives for situations like this. If we could target this to just the portfolio pattern, I would say 100% go for it, but this feels like a broken experience in multiple patterns to me. This is my home template with just one image on one post, and I think it looks less broken than if we had the white spaces for the other 2:

Screenshot 2023-09-25 at 12 54 41

scruffian:
What do you think about updating the query block so that we can filter out posts that don't have a featured image?

carolinan:
@scruffian I think it is too late in the release cycle, even if the code was completed it would not be tested well enough.

carolinan:
This is why it would be great to start the work on the bundled theme earlier in the year, because we do find things that can't be done yet, that needs to be added or updated in core, that needs time to be tested as part of Gutenberg before it is added to core.

MaggieCabrera:
> This is why it would be great to start the work on the bundled theme earlier in the year, because we do find things that can't be done yet, that needs to be added or updated in core, that needs time to be tested as part of Gutenberg before it is added to core.

I still think that it's not a bad thing if we push for something like this for 6.5. Even if the change doesn't make it at the same time as the theme, it will still be there. Default theme-related changes shouldn't only be relevant on the release the theme ships into core

carolinan:
There is also the option to look for the first image in the content.

MaggieCabrera:
> There is also the option to look for the first image in the content.

We still risk that there might be no image there

beafialho:
I don't think this is needed, to be honest. We provide three archive templates in the theme: the **default** one, the **photographer** (just featured images) and **writer** which looks perfectly without featured images, enough options for users to pick the one that works best for them.

I second @MaggieCabrera, I think the white box makes it look like a bug.

carolinan:
I think these templates are unusable as is, and I think many users will struggle, but it is not my call.

PR #500, opened 2023-09-25
Remove screen-reader-text class from headings

carolinan:
**Description**
The screen-reader-text class completely hides the block in the editor, so this can not be used to hide the block on the front.
This PR removes the class, **and does not attempt to hide the text again,** so the text will be visible.

**Screenshots**
Index:
Index with visible H1

Single post comment area:
Visible heading above comments area

**Testing Instructions**
Test each pattern and template that had the hidden headings; comments area, index, portfolio index.


carolinan:
This also fixes the block validation error.

luminuu (approved):
LGTM

PR #501, opened 2023-09-25
Add aria labels to the placeholder footer navigation

carolinan:
**Description**
Adds aria labels to the placeholder navigation blocks in the footer.
This helps screen reader users identify the different menus.
Requirement: https://make.wordpress.org/themes/handbook/review/accessibility/required/#aria-landmark-roles

https://github.com/WordPress/twentytwentyfour/assets/7422055/e59a0e50-34a8-4e15-85d4-98caf9853e66

**Testing Instructions**
Go to the front of the website and view the source of the footer. Confirm that the three `<nav>` elements have an aria-label that matches the heading.

### DIFF
--- a/patterns/footer.php
+++ b/patterns/footer.php
@@ -33,7 +33,7 @@
 <!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|10"}},"layout":{"type":"flex","orientation":"vertical"}} -->
 <div class="wp-block-group">
 
-<!-- wp:navigation {"overlayMenu":"never","layout":{"type":"flex","orientation":"vertical"},"style":{"typography":{"fontStyle":"normal","fontWeight":"400"}},"fontSize":"small"} -->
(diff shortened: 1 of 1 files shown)


luminuu:
If a user changes just the content, so the headline and the links, does the aria label stay as is or does it get updated too?

carolinan:
It stays.

But if they add a real **menu** to the navigation block it is updated to use the menu name.

carolinan:
I was partially wrong. If the links are updated through the UI then the user is asked to create a *menu*, and this saves a new aria label, with the menu name.

If they update the *code* it stays.

luminuu:
Hm, so what is the better option? Having no aria label at all or having one that maybe gets updated, but if not, is wrong too?

carolinan:
Hopefully, if someone is able to edit the code directly, they will also see the label.
Without being able to identify the different navs, it won't pass the accessibility-ready requirements.

luminuu (approved):
I'm merging this in now. 

Is there an issue somewhere at the Gutenberg repo to raise the issue that the aria label might not be updated when a user is just changing the content? Working in the code editor is probably not something the majority of regular users will do.

carolinan:
I have not created an issue for it.

Issue #502, opened 2023-09-25
Javascript warning in the home template in the Site Editor


carolinan:
**Description**
In the Site Editor, there is a JavaScript warning when you view the default home template.

`Warning: Encountered two children with the same key, `post-meta`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`

I did some manual testing, and this seems to be because one of the patterns uses the post-meta template part, twice,
but more testing is always helpful.

**Opinion:** Reading the warning, this seems to be intentional. I don't know the background to it, but if there are good reasons why a template part can not be used twice in the same template, then it should be solved in the theme.

**Step-by-step reproduction instructions**
Open the home template in the Site Editor and see the JS warning in the browser console.
In posts-featured.php, remove one of the two post-meta template parts.
Confirm that the warning disappears.

carolinan:
So, I manually created a custom template through the UI, and added the post meta template part twice. The warning appeared. (though not until I had saved and then refreshed the editor a few times).

carolinan:
If this is a React warning I don't see how we can solve it without removing one of the parts?

MaggieCabrera:
I think this is a react warning, but the meta is showing twice. I think it should be fixed, but I'm not 100% confident, so I'll open a GB PR and ask for help reviewing

MaggieCabrera:
https://github.com/WordPress/gutenberg/pull/54861

carolinan:
I believe this can be closed?

MaggieCabrera:
Yes, this was fixed upstream

PR #506, opened 2023-09-26
Remove Link on Post Meta Date

### LINKED ISSUE #504: Single posts links to themselves
**Description**
In the single post view, the post date is linked to the current post. 

Open a single post on the front.
Confirm that the date is linked to the post.

**Expected behavior**
For a page to not link to itself.

bobmatyas:
**Description**

Removes date link from post meta template. Fixes https://github.com/WordPress/twentytwentyfour/issues/504

**Screenshots**

After:

Screenshot 2023-09-25 at 9 03 18 PM

**Testing Instructions**

1. Checkout PR and test.

**Contributors**


carolinan:
Hi
This removes the link also from archives. So when a post has no title, there is no way to reach that post from the archive.

bobmatyas:
Ahh shoot, I didn't test that scenario. I'll look at adding a conditional check and submitting a revised PR tomorrow.

luminuu:
I'm not sure if this makes sense to put in a conditional there or if this should be an option on the post date block itself, which then has to be raised in the Gutenberg repo.

carolinan:
There are different ways to solve this; these are my suggestions for short-term solutions, either:
- Use two different parts for the post meta. (Like one for archives, one for single)
- Remove the link from all post dates and add a read more block to the queries.

The reason why I recommend the first is because it could also be the best solution for this issue: https://github.com/WordPress/twentytwentyfour/issues/502

Long term, I am not sure that adding another condition to the control for the block would be best because it may be confusing to some users; maybe under advanced.

MaggieCabrera:
> There are different ways to solve this; these are my suggestions for short-term solutions, either:
> 
>     * Use two different parts for the post meta. (Like one for archives, one for single)
> 
>     * Remove the link from all posts and add a read more block to the queries.
> 
> 
> The reason why I recommend the first is because it could also be the best solution for this issue: #502
> 
> Long term, I am not sure that adding another condition to the control for the block would be best because it may be confusing to some users; maybe under advanced.

On the other hand, having a read more link seems more intuitive than having the date be the link. I'd solve the post-meta problem using a pattern and a template part on that offending template. Having two identical template parts seems confusing and the user won't know right away which templates use which without having to check each one of them

carolinan:
@MaggieCabrera @beafialho Is there a new design that includes a read more / continue reading?

beafialho:
Yes, here's the design:

Captura de ecrã 2023-10-03, às 12 19 49

`Read More →` uses the same font size as post meta.

It can be seen on Figma [here](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=492-1806&mode=design&t=Tcut1PnuNtj2kWTw-4).

carolinan:
So the meta is now above the excerpt, not below?

beafialho:
It's always been, no? As seen [here](http://2024.wordpress.net/index.php/writer-home/).

carolinan:
Only one some of the post lists.

luminuu:
Carolina is refering to this pattern that is used with the business/entrepreneur home page, where the post meta is coming after the excerpt:

CleanShot 2023-10-04 at 13 30 02@2x

MaggieCabrera:
The post meta is used in a lot of places, I think we can keep everything where it is and implement the read more link like on the screenshot Bea provided without altering any of the other blocks. The arrow can be implemented in a similar way that I did with the block style for nav link items (it's a different arrow there, but the implementation should be very similar if not the same)

richtabor:
I'm going to close this. I agree, we should keep a consistent post-meta across the board; as we're using it in template part fashion. Makes it much easier to change from the user's perspective wherever it's used.

PR #508, opened 2023-09-26
Audit usage of locked blocks

### LINKED ISSUE #400: Audit block locking
**Description**
Some blocks have `lock":{"move":false,"remove":false}` which doesn't do anything.
This can be removed. The post content blocks should probably stay locked (remove: true) though.

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/400

Removed usage of `lock` on blocks that are not the post content. Allowed post-content block to be removed

### DIFF
--- a/templates/page-wide.html
+++ b/templates/page-wide.html
@@ -16,7 +16,7 @@
 <!-- /wp:column -->
 
 <!-- wp:column -->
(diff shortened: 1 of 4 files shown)


luminuu (approved):
LGTM

PR #517, opened 2023-09-27
Remove break causing space at beginning of credit

### LINKED ISSUE #512: All footers: Remove space in "Designed with WordPress"
There's an unintentional space behind the words "Designed with WordPress" in every footer, which should be removed.

https://github.com/WordPress/twentytwentyfour/assets/65220155/794f6fd8-09cc-4eff-87e9-dd30c2dbbbd7

jordesign:
**Description**

Fixes #512 by removing the space before ' Designed'

**Screenshots**

Screenshot 2023-09-27 at 12 29 47 pm

**Questions**

The only way I found to do this was by removing the line breaks between the opening paragraph tag and the opening PHP tag. This seems like it won't fit with the coding styleguide - but I'm not sure where the space was coming from otherwise.

### DIFF
--- a/patterns/footer-portfolio.php
+++ b/patterns/footer-portfolio.php
@@ -82,8 +82,7 @@
 			</div>
 			<!-- /wp:group -->
 			<!-- wp:paragraph {"fontSize":"small"} -->
-			<p class="has-small-font-size">
-				<?php
+			<p class="has-small-font-size"><?php
 					/* Translators: WordPress link. */
 					$wordpress_link = '<a href="' . esc_url( __( 'https://wordpress.org', 'twentytwentyfour' ) ) . '" rel="nofollow">WordPress</a>';
 					echo sprintf(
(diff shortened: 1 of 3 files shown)


MaggieCabrera:
The linter is complaining about this >< I'm not sure how to go about this one

jordesign:
It seems weird that it is adding a space _inside_ the paragraph, just from those line breaks in the template

luminuu:
Personally I'd prefer to have the linter not failing. I know this isn't the best option, but visually on an unedited site, it's not visible:

CleanShot 2023-09-27 at 15 49 12@2x

Only in the editor the space is there.

jordesign:
> Personally I'd prefer to have the linter not failing. I know this isn't the best option, but visually on an unedited site, it's not visible:


That seems like the better experience overall - I'd agree.
I think ok to just close this PR then?

luminuu:
Yep, I think we can close this PR.

PR #522, opened 2023-09-27
Avoid empty headings with asterisk

MaggieCabrera:
**Description**

I've been chatting with @scruffian about the current implementation of the asterisks and he supports that https://github.com/WordPress/twentytwentyfour/pull/449 is a better solution because with the solution on the heading block users can introduce empty hedings just to make use of the asterisk. 

This PR aims to partially fix that by hiding the asterisk on empty heading blocks. The moz prefixed rule makes it work on firefox when the user introduces whitespace in the heading too (chrome doesn't have a solution for that yet).

I would like to keep the discussion open if we feel like https://github.com/WordPress/twentytwentyfour/pull/449 is a better option for this problem. I hesitate because I think that block style in the sapcer is not a good precedent for the default theme to set regarding block styles, it feels like bending a little too much what the core block does, but I see the point that the heading solution isn't perfect either so I would like to hear what other people have to say about this.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -129,6 +129,14 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 					clip-path: path('M11.93.684v8.039l5.633-5.633 1.216 1.23-5.66 5.66h8.04v1.737H13.2l5.701 5.701-1.23 1.23-5.742-5.742V21h-1.737v-8.094l-5.77 5.77-1.23-1.217 5.743-5.742H.842V9.98h8.162l-5.701-5.7 1.23-1.231 5.66 5.66V.684h1.737Z');
 					display: block;
 				}
+
+				.is-style-asterisk:empty:before {
+					content: none;
+				}
+				
+				.is-style-asterisk:-moz-only-whitespace:before {
+					content: none;
+				}
 				
 				.is-style-asterisk.has-text-align-center:before {
 					margin: 0 auto;
--- a/functions.php
+++ b/functions.php
@@ -130,6 +130,7 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 					display: block;
 				}
 
+				/* Hide the asterisk if the heading has no content, to avoid using empty headings to display the asterisk only, which is an A11Y issue */
 				.is-style-asterisk:empty:before {
 					content: none;
 				}


scruffian:
IMO the asterisk is a decorative element with no semantic meaning and therefore the spacer group seems like the best semantic fit for it. I think we should should consider it a "decorative" spacer, rather than an empty spacer.

luminuu:
From a user perspective, I think adding a block style to a spacer block blurs the functionality with a separator block. How is a user supposed to know what the difference is between a spacer block and the separator block? Because the separator block does have visual differences besides just a simple line (wide line, three dots). I'm aware of the difference that one is announced in screen readers and the other isn't. 

From a developer perspective, I agree that in terms of accessibility the spacer makes more sense than the heading or the separator. But how do you communicate this to the average user? I'm not so sure about this.

carolinan (approved):

functions.php
```diff
@@ -129,6 +129,14 @@ function twentytwentyfour_block_styles() {
 					clip-path: path('M11.93.684v8.039l5.633-5.633 1.216 1.23-5.66 5.66h8.04v1.737H13.2l5.701 5.701-1.23 1.23-5.742-5.742V21h-1.737v-8.094l-5.77 5.77-1.23-1.217 5.743-5.742H.842V9.98h8.162l-5.701-5.7 1.23-1.231 5.66 5.66V.684h1.737Z');
 					display: block;
 				}
```
carolinan, line 131: Perhaps include a comment about why this is needed?

jeryj:
> I think adding a block style to a spacer block blurs the functionality with a separator block. 

I honestly didn't know about the difference between a separator and a spacer until working on this asterisk issue. I agree that it does blur the lines of the functionality between them. 

If we used the spacer idea, would it get used as a pattern? If so, my guess is that the average user would not notice that it was used within a spacer and not a separator, nor would they care that the technical difference between them is blurred. I think they would copy the existing block or select the pattern, likely never using the spacer directly. However, for users who do know Gutenberg well, they may look for the asterisk within the separator block, as that does make the most sense. 

I feel like the best solution would be if we could have a non-semantic option within the separator block. I bet plenty of people are using the separator block decoratively on other themes. Maybe this is justification for a change within core to add an option to output the separator block as a `<div>`?

Or, is there a way to filter the save function of the Separator on the PHP side, so TT4 saves the separator as a `<div>` instead of an `<hr>` if you have the asterisk option selected?

scruffian:
We can't modify the save function of the separator block on the PHP side as it's not a dynamic block. It might be possible to modify the output of the block by modifying it's save function. I had a quick look at that and it seems non-trivial.

carolinan:
Could it be an option to change the element under the advanced panel of the separator?

luminuu:
> Could it be an option to change the element under the advanced panel of the separator?

I like that idea. It brings us closer to what a user can expect from a separator block.

The closest issues I could find in the Gutenberg repo are:
https://github.com/WordPress/gutenberg/issues/34743
https://github.com/WordPress/gutenberg/issues/16483

However they do not include an option to change the element. Would be great if we can follow up with this, for now I'll merge in this PR for the next Beta.

functions.php
luminuu: Done

luminuu:
Update: added an issue to the Gutenberg repo: https://github.com/WordPress/gutenberg/issues/54976

Issue #520, opened 2023-09-27
Buttons: Remove border that gets visible on hover


beafialho:
Currently when we hover buttons, the border gets visible. Let's remove it so the states look like the design.

**Currently:**

https://github.com/WordPress/twentytwentyfour/assets/65220155/0a521ae3-d862-470d-a72e-3cfe5a8a1409

**Intended:**

https://github.com/WordPress/twentytwentyfour/assets/65220155/2a6de00b-c209-4a4c-8253-197d9794d764

MaggieCabrera:
The border needs to be there so the size of the normal button is the same as the outlined one, but the color of the border on hover does need fixing so it matches the desired behaviour

PR #526, opened 2023-09-27
resolve border issue when button is hovered

NidhiDhandhukiya74:
**Description**
Resolve the issue described in the issue :- #520

### DIFF
--- a/assets/css/button-outline.css
+++ b/assets/css/button-outline.css
@@ -4,3 +4,7 @@
 	color: var(--wp--preset--color--base);
 	border-color: var(--wp--preset--color--contrast-2);
 }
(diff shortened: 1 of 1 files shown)


assets/css/button-outline.css
```diff
@@ -4,3 +4,7 @@
 	color: var(--wp--preset--color--base);
 	border-color: var(--wp--preset--color--contrast-2);
 }
+.wp-block-button.is-style-fill
```
MaggieCabrera, line 7: Can this be done in theme.json instead? button-outline.css will only load when that block style is being used on the page.

luminuu:
Closed in favor of #542

PR #530, opened 2023-09-28
Enable lightbox by default

### LINKED ISSUE #529: Set default image behavior to expand on click (lightbox)
To showcase the new lightbox functionality in core, I propose we set lightbox (Expand on click) on for images. 

- [ ] Update theme.json to reflect lightbox global setting ([source](https://github.com/WordPress/gutenberg/blob/trunk/lib/theme.json#L276-L280))
- [ ] Update patterns that should have lightbox off, like clients

richtabor:
**Description**

Closes #529 by enabling lightbox on by default for images.

**Testing Instructions**

1. Activate the theme.
2. Add a page.
3. Add an image. 
4. See the "Expand on click" control active by default.
5. Add the "Clients" pattern.
6. Add an image to one of the placeholders. 
7. See "Expand on click" turned off.


richtabor:
Going to close this as not-planned. The lightbox feature needs a bit more attention before we should enable it by default out-of-the-box. This is a good PR to showcase the _how_ though, for future reference.

PR #531, opened 2023-09-29
Replace the RSVP patten with the full page RSVP landing page pattern

### LINKED ISSUE #491: Full page pattern for RSVP
Bea marked this pattern as full page on the demo site and I agree that this makes sense. Let's treat this one the same way we did the newsletter landing one.

http://2024.wordpress.net/index.php/rsvp/

carolinan:
**Description**
This PR replaces the RSVP pattern with the RSVP landing page.

Closes https://github.com/WordPress/twentytwentyfour/issues/491

1. The file has been renamed, and the file-header updated, but the markup remains the same except for the large text.
2. The paragraph has been replaced with an H1 to improve the accessibility (which is why I labeled this as a bug fix).

64 local_89-2_

**Testing Instructions**
This pattern works best with the "no title" page template:

1. Create a new page. 
2. In the pattern selection modal, select the RSVP pattern.
3. Do not add a title. 
4. Select the "No title" page template.
5. Save and view the page on the front.

### DIFF
--- a/patterns/page-07-rsvp-landing.php
+++ b/patterns/page-07-rsvp-landing.php
@@ -1,9 +1,12 @@
 <?php
 /**
- * Title: RSVP
- * Slug: twentytwentyfour/rsvp
- * Categories: call-to-action, banner, featured
- * Viewport width: 1400
+ * Title: RSVP Landing Page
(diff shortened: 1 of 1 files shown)


luminuu (approved):
LGTM

PR #536, opened 2023-10-02
Change from main to aside

sumisubedi:
The Sidebar should have "tagName":"aside, " which is missing on the template with the sidebar. This PR will add that.


MaggieCabrera (approved):

MaggieCabrera:
This works, but we have other instances of the sidebar where I think this is needed as well, can you update the PR adding those, please?

sumisubedi:
@MaggieCabrera Would you please check once? I added on two more where the sidebar is present.

MaggieCabrera (approved):

PR #540, opened 2023-10-02
Removed sticky behavior from header

### LINKED ISSUE #524: Don't make the header sticky by default
**Description**

it's a known fact that fixed / sticky headers or top bars may introduce usability and accessibility issues. While testing the footnotes block in the context of https://github.com/WordPress/gutenberg/pull/54843 I noticed that the sticky header in the theme makes the user experience less than ideal.

Specifically, following a footnote link and clicking the 'jump back' link in the footnote does make the page scroll as expected _but_ the page frgment that should scroll into view is always hidden behind the fixed header.

This is a very confusing experience for all users (it happens also when clicknw with mouse) and even more confusing for keyboard users. 

Since the sticky header height can't be predicted, as it actually depends on the amount of links in the navigation, the only option I can think of is to change the default setting. The header should not be sticky by default. That would provide users with a default user experience that is less than ideal.

**Expected behavior**

Focused links and controls to be visible.

**Screenshots**

Animated GIF to illustrate the footnotes behavior:

tt4 sticky header

**Environment info**

**Additional context**

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/524

Since Bea is happy with this change and it's an a11y issue, I have no issues with removing the stickiness from the header.

### DIFF
--- a/patterns/template-archive-portfolio.php
+++ b/patterns/template-archive-portfolio.php
@@ -8,11 +8,7 @@
  */
 ?>
 
-<!-- wp:group {"layout":{"type":"default"}} -->
-<div class="wp-block-group">
-	<!-- wp:template-part {"slug":"header"} /-->
-</div>
-<!-- /wp:group -->
+<!-- wp:template-part {"slug":"header"} /-->
 
 <!-- wp:group {"tagName":"main","align":"full","layout":{"type":"constrained"}} -->
 <main class="wp-block-group alignfull">
--- a/patterns/template-archive-portfolio.php
+++ b/patterns/template-archive-portfolio.php
@@ -8,7 +8,7 @@
  */
 ?>
 
-<!-- wp:group {"style":{"position":{"type":"sticky","top":"0px"}},"layout":{"type":"default"}} -->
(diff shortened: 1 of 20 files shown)


patterns/template-archive-portfolio.php
```diff
@@ -8,7 +8,7 @@
  */
 ?>
 
-<!-- wp:group {"style":{"position":{"type":"sticky","top":"0px"}},"layout":{"type":"default"}} -->
+<!-- wp:group {"layout":{"type":"default"}} -->
 <div class="wp-block-group">
```
carolinan: Is the group still needed? I thought it was only added for the sticky option.

MaggieCabrera: I think we need it for the layout, but we can test

luminuu: I just tested, it can be removed and still works fine. I'm working on a commit now.

luminuu (approved):
Updated to remove the additional group on all templates, LGTM!

PR #544, opened 2023-10-02
Sync with core after Beta 2

MaggieCabrera:
**Description**

This PR syncs with core so we have the same code in both repos


luminuu (approved):
LGTM

PR #547, opened 2023-10-02
Update version on readme

huzaifaalmesbah:
i update on readme.txt Requires at least: 6.2 to 6.4 for match style.css and readme.txt [check](https://github.com/WordPress/twentytwentyfour/blob/be827b8e3a5932191511c033d73a2898e110bed5/style.css#L7)


kafleg (approved):
LGTM

Issue #533, opened 2023-09-29
Pulloquote Block border not displaying.


viralsampat-multidots:
**Description**
Hello Team,

I have worked on theme "**Twenty Twenty Four**" and found that the "**Pulloquote**" block border is not displaying when we trying to apply it from block setting.

**Step-by-step reproduction instructions**
1. Go to page Type / to choose a block
2. Select "**Pullquote**" block
3. Add dummy text
4. Click on block setting and try to set border.

**Expected behavior**

![CleanShot 2023-09-29 at 15 25 28@2x](https://github.com/WordPress/twentytwentyfour/assets/53076293/5a46a1e6-2244-4f25-b167-3288eb433460)

**Screenshots**
Issue Video: (https://share.cleanshot.com/N0Phrr0rJ0HwVW5MN16T)

**Back-end:**
![CleanShot 2023-09-29 at 15 36 43@2x](https://github.com/WordPress/twentytwentyfour/assets/53076293/5527f8f7-37d2-45d0-b047-d790b5dcf93d)

**Front-end**
![CleanShot 2023-09-29 at 15 36 17@2x](https://github.com/WordPress/twentytwentyfour/assets/53076293/4e22e09b-026c-432e-b1c5-92b5fdbf9dec)

**Environment info**
Device: Macbook M1
OS: 13.5.2 (22G91)
Browser: Google chrome
Version: 117.0.5938.132 (Official Build) (arm64)
WordPress version: 6.3.1, Gutenberg 16.7.0, Theme active: Twenty Twenty-Four,

zakariabinsaifullah:
I have fixed the issues and added PR #535. I am waiting for the merge.

luminuu:
Done with #550

PR #550, opened 2023-10-03
Fixed Pulloquote Block border issue

viralsampat-multidots:
**Description**

Hello Team,

I have worked on issue #533 and resolved it from my end. Here, I have added my PR.

**Screenshots**

**Back-end**
backend

**Front-end**
frontend

**Testing Instructions**

- Go to page Type / to choose a block
- Select "Pullquote" block
- Add dummy text
- Click on the block setting and try to set the border.

**Contributors**

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -612,23 +612,9 @@
 			},
 			"core/pullquote": {
 				"border": {
-					"bottom": {
-						"style": "solid",
-						"width": "1px"
-					},
-					"left": {
-						"style": "solid",
-						"width": "1px"
-					},
 					"radius": "var(--wp--preset--spacing--20)",
-					"right": {
-						"style": "solid",
-						"width": "1px"
-					},
-					"top": {
-						"style": "solid",
-						"width": "1px"
-					}
+					"style": "solid",
+					"width": "1px"
 				},
 				"elements": {
 					"cite": {
--- a/theme.json
+++ b/theme.json
@@ -612,8 +612,7 @@
 			},
 			"core/pullquote": {
 				"border": {
-					"radius": "var(--wp--preset--spacing--20)",
-					"style": "solid"
+					"radius": "var(--wp--preset--spacing--20)"
 				},
 				"elements": {
 					"cite": {
--- a/theme.json
+++ b/theme.json
@@ -613,21 +613,21 @@
 			"core/pullquote": {
 				"border": {
 					"bottom": {
-						"style": "none",
-						"width": "0px"
+						"style": "solid",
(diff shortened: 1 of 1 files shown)


theme.json
```diff
@@ -613,21 +613,21 @@
 			"core/pullquote": {
 				"border": {
 					"bottom": {
-						"style": "none",
-						"width": "0px"
+						"style": "solid",
```
luminuu: In order to make the borders work again and keep the layout as in the design, the `bottom`, `left`, `right` and `top` attributes should be completely removed. This way there's no border when using the block as default, and borders can be added if the block settings are updated.

Please remove the mentioned attributes, but keep the `radius` attribute.

luminuu (changes requested):

theme.json
```diff
@@ -612,23 +612,9 @@
...
 					"radius": "var(--wp--preset--spacing--20)",
-					"right": {
-						"style": "none",
-						"width": "0px"
-					},
-					"top": {
-						"style": "none",
-						"width": "0px"
-					}
+					"style": "solid",
```
luminuu: This does not need to be added, as the default design of the pullquote block does not have a border. Please remove these two lines.

luminuu (changes requested):

theme.json
viralsampat-multidots: Hello @luminuu 

I have added my new PR.

Please review it and let me know your feedback for the same.

Thanks,

theme.json
viralsampat-multidots: Hi @luminuu 

Done,

Please check it

Thanks,

theme.json
```diff
@@ -612,23 +612,8 @@
...
 					"radius": "var(--wp--preset--spacing--20)",
-					"right": {
-						"style": "none",
-						"width": "0px"
-					},
-					"top": {
-						"style": "none",
-						"width": "0px"
-					}
+					"style": "solid"
```
luminuu: Please remove the `"style": "solid"` too, as it creates a border that is not in the design:

CleanShot 2023-10-04 at 13 07 38@2x

luminuu (changes requested):

theme.json
viralsampat-multidots: Hi @luminuu 

I have updated my PR and removed the above-mentioned line.

Please review it.

Thanks,

luminuu (approved):
Tested. By default, the pullquote has no border as intended in the design, and borders can be added via the block settings if needed.

Issue #552, opened 2023-10-03
Details block having issue when Arrow Icon has chosen and border applied.


NidhiDhandhukiya74:
**Description**

Details Details block having issue when Arrow Icon has chosen and border applied.

**Step-by-step reproduction instructions**

1. Take Details block.
2. Apply some background color to check the proper difference
3. Use Border.

**Expected behavior**

Border needs to be applied all 4 sides.
But border-bottom is not looking same.

**Video**

Video URL :- https://share.cleanshot.com/m80xKYfPg92fyQJK8nsH

**Environment info**

Mac OS
Chrome Browser
Wordpress 6.3.1
Gutenberg - 16.7.0

PR #553, opened 2023-10-03
Fix bottom border issue with arrow style

NidhiDhandhukiya74:
**Description**
Resolve the mentioned issue of :- #552

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -52,7 +52,6 @@ if ( ! function_exists( 'twentytwentyfour_block_styles' ) ) :
 				.is-style-arrow-icon-details {
 					padding-top: var(--wp--preset--spacing--10);
 					padding-bottom: var(--wp--preset--spacing--10);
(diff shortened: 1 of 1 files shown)


functions.php
```diff
@@ -52,7 +52,6 @@ function twentytwentyfour_block_styles() {
 				.is-style-arrow-icon-details {
 					padding-top: var(--wp--preset--spacing--10);
 					padding-bottom: var(--wp--preset--spacing--10);
-					border-bottom: 1px solid rgba(255, 255, 255, 0.20);
```
MaggieCabrera, line 55: I think instead of removing, we probably need to use something like currentColor

MaggieCabrera: also, I think this is not really an issue since you can choose the border color in the color picker and it will override this CSS.

MaggieCabrera:
I'm closing this one since it's something the user can override:

Screenshot 2023-10-04 at 10 16 54

PR #555, opened 2023-10-04
Remove setCascadingProperties from nav block

### LINKED ISSUE #554: Remove deprecated attribute from Navigation block
**Description**

In the [Navigation block](https://github.com/WordPress/twentytwentyfour/blob/39b80cd814cbdfa68fe5599753fa6710d656f2f2/parts/header.html#L11) included in the header, there is a `setCascadingProperties` attribute that is no longer in use. It was part of the initial implementation of flex layout in the Navigation block, but was [removed](https://github.com/WordPress/gutenberg/pull/37473) before landing in core.
I only recently realised that, because the removal was already during the Beta stage, that attribute was included in the TT2 markup, and so must have been copied into TT3 and now into TT4 😅 

It would be great if we could finally break the cycle! Removing the attribute will have no effect apart from cleaning up the markup, because that initial implementation was scrapped.

MaggieCabrera:
**Description**

Removes the deprecated attribute and closes https://github.com/WordPress/twentytwentyfour/issues/554

The navigation should still work as intended

### DIFF
--- a/parts/header.html
+++ b/parts/header.html
@@ -8,6 +8,6 @@
 	<!-- /wp:group --></div>
 	<!-- /wp:group -->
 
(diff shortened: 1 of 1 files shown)


luminuu (approved):
Tested, did not have any issues with the navigation.

PR #556, opened 2023-10-04
Remove static colors in favor of presets

### LINKED ISSUE #417: Colors in the stylesheet
**Description**

The theme declares a few colors in some of the block styles like the arrows and the post tags. Those color won't work well with style variations so we need to use some of the presets defined by the theme.json. Let's change the hex codes with CSS variables

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/417

This PR removes the static colors from the FAQ pattern and associated block style in favor of presets. The color is not the exact same as the figma but it's quite close. This makes it work on the style variations too.

**Screenshots**

Screenshot 2023-10-04 at 10 15 00

Screenshot 2023-10-04 at 10 14 54


luminuu (approved):
LGTM

PR #557, opened 2023-10-04
Single posts: fix padding on mobile

### LINKED ISSUE #421: Single post: Content reaches the edge of the browser on small widths
**Description**
In single posts, the content reaches the edge of the browser. Even content that is not set to full width itself.
The content starts moving closer to the edge at ca 740px.

**Expected behavior**
I expect there to be spacing between the content and the edge of the browser, unless the individual block inside the post content is set to full width.

**Screenshots**
Content positioning on small screen

**Environment info**

 - Device: Desktop
 - OS: macOS
 - Browser: Chrome
 - Version:  117.0.5938.88 
 
 6.4 current trunk, Gutenberg current trunk

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/421

This PR fixes the markup of this template so we have padding on the sides of the post content while maintaining the ability to add wide and full width aligned blocks on desktop

**Screenshots**

Screenshot 2023-10-04 at 10 28 27
Screenshot 2023-10-04 at 10 28 19

### DIFF
--- a/templates/single.html
+++ b/templates/single.html
@@ -1,6 +1,6 @@
 <!-- wp:template-part {"slug":"header"} /-->
 
-<!-- wp:group {"tagName":"main","align":"full","layout":{"type":"constrained"}} -->
+<!-- wp:group {"tagName":"main","align":"full"} -->
 <main class="wp-block-group alignfull"><!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|50"},"margin":{"bottom":"var:preset|spacing|40"}}},"layout":{"type":"constrained"}} -->
 <div class="wp-block-group" style="margin-bottom:var(--wp--preset--spacing--40);padding-top:var(--wp--preset--spacing--50)"><!-- wp:post-featured-image {"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|40"}}}} /-->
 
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
@richtabor unrelated to this PR but the screenshots speak for themselves. Should we reconsider the way the smaller spacing presets scale on mobile? You can see how the FAQ items go from having a 1rem padding on desktop to 3px on mobile (preset 10)

richtabor:
> Should we reconsider the way the smaller spacing presets scale on mobile? 

We can test changing just that smallest preset to a fixed value. Would need to make sure it's solid anywhere it's employed though.

templates/single.html
```diff
@@ -1,6 +1,6 @@
 <!-- wp:template-part {"slug":"header"} /-->
 
-<!-- wp:group {"tagName":"main","align":"full","layout":{"type":"constrained"}} -->
+<!-- wp:group {"tagName":"main","align":"full"} -->
 <main class="wp-block-group alignfull"><!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|50"},"margin":{"bottom":"var:preset|spacing|40"}}},"layout":{"type":"constrained"}} -->
```
webmandesign, line 4:
```html
<!-- wp:group {"tagName":"main","align":"full"} -->
<main class="wp-block-group alignfull">
```

can be further improved to 

```html
<!-- wp:group {"tagName":"main"} -->
<main class="wp-block-group">
```

templates/single.html
```diff
@@ -13,16 +13,10 @@
 <!-- /wp:group --></div>
 <!-- /wp:group -->
 
-<!-- wp:group {"align":"full","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|40"}}},"layout":{"type":"default"}} -->
-<div class="wp-block-group alignfull" style="margin-bottom:var(--wp--preset--spacing--40)">
+<!-- wp:post-content {"lock":{"move":false,"remove":true},"align":"full","layout":{"type":"constrained"}} /-->
```
webmandesign, line 16:
```html
<!-- wp:post-content {"lock":{"move":false,"remove":true},"align":"full","layout":{"type":"constrained"}} /-->
```

can be further improved to

```html
<!-- wp:post-content {"lock":{"move":false,"remove":true},"layout":{"type":"constrained"}} /-->
```

luminuu:
@webmandesign Why do you want to have the `align: full` removed?

templates/single.html
MaggieCabrera: Have you tested if this change still allows for full width blocks inside the post content?

luminuu (approved):

templates/single.html
webmandesign: Yes. It does work on WP 6.4-beta2 with Gutenberg 16.7.1.

The parent here is `.wp-site-blocks` which has no layout styles. Full/wide alignment only works within a constrained layout container, so having it here is redundant and editor does not even display controls to set/change the alignment.

templates/single.html
webmandesign: This works on WP 6.4-beta2 with Gutenberg 16.7.1.

The parent here is `main.wp-block-group` with default (not constrained) layout.

Full/wide alignment only works within a constrained layout container, so having it here is redundant and editor does not even display controls to set/change the alignment.

webmandesign:
@luminuu Please see my additional comments for explanation. Basically, the `align: full` is redundant when parent group has no constrained layout.

PR #558, opened 2023-10-04
Rust varition: fixes for missing colors

### LINKED ISSUE #480: Button Hover text color issue in Rust Style Variation
## **Description**
In Rust Style Default Button Text Color is not visible & in Outline Button Hover text Color  is not visible.

### **On Default Button hover:**
rust-style

### **On button outline Hover:**
outline-button-on-hover

MaggieCabrera:
**Description**

Rust only defines 3 colors and we have a few places where the undefined colors are used. This PR assigns new colors in those instances so we keep everything visible and contrasting.

Closes https://github.com/WordPress/twentytwentyfour/issues/480


luminuu (approved):
LGTM

Issue #3, opened 2023-08-24
Meet WCAG AA level accessibility requirements


luminuu:
2024 should be WCAG AA level compliant. What is needed to achieve this? Let's use this issue to discuss and track.

Some resources:
- Checklist by A11Y Project: https://www.a11yproject.com/checklist/
- WebAIM: https://webaim.org/standards/wcag/checklist
- Theme Handbook: https://make.wordpress.org/themes/handbook/review/accessibility/required/

For reference, here's the same discussion for Twenty Twenty-Two: https://github.com/WordPress/twentytwentytwo/issues/82

carolinan:
### Result of the accessibility-ready review

[Requirements](https://make.wordpress.org/themes/handbook/review/accessibility/required/)

Keyboard Navigation -pass, (Besides known issues reported for the navigation block in the Gutenberg GitHub repository)
Controls -pass
Skip Links -pass except for the portfolio home template
Forms -pass.

**Headings**
H1 is missing on the default home template
H1 is missing on the writer home template
H1 is missing from the portfolio home template

The content of the portfolio page pattern is difficult to read because it has no headings. (This is the portfolio pattern that shows up when you create a new single page)

**ARIA Landmark Roles**
The sidebar should have its own landmark (aside)
Single with sidebar: the main is in the wrong place, it does not wrap the featured image, title or post meta.
Main is missing from the portfolio home template

**Content Links**
Some links in the sidebar does not change visually when you :hover over them
(Perhaps the underline could be removed on hover).

Repetitive Link Text -pass

Contrasts
Needs more careful testing
Milky: The default link color has too low contrast against the background


Not tested yet:
Images
Media
Screen Reader Text
Not Allowed

kafleg:
Contrast

Onyx Style - Very low contrast between text and background colors found in the home page testimonial section.

carolinan:
Continued: 

Images -Pass.   _For the lightbox PR to be merged, the lightbox accessibility issues in core/Gutenberg must first be fixed._
Media -pass
Screen Reader Text -pass
Not Allowed -pass

carolinan:
If the linked post date block is replaced with a "read more", the "repetitive link text" requirement needs to be considered and a new review should be done.

kafleg:
Contrast error found in Milky style.
![image](https://github.com/WordPress/twentytwentyfour/assets/13062231/f053f869-35a9-40cf-8e22-7736f505bf4a)

[...]

PR #563, opened 2023-10-04
Add missing H1 headings

carolinan:
**Description**
For https://github.com/WordPress/twentytwentyfour/issues/3

The PR solves the following:

H1 is missing on the default home template
H1 is missing on the writer home template
H1 is missing from the portfolio home template

In some templates this means replacing paragraphs with H1 headings, which means that I removed the font family and in one case, the line height. Kindly review that the text still matches the wanted design.

**Testing Instructions**
View the templates listed above.  Confirm that the template has an H1 heading.
Confirm that the style of the text is correct.


luminuu (approved):
Tested, LGTM

Issue #566, opened 2023-10-04
Archive template (No results): Spacing and copy adjustment


beafialho:
The spacing below the title and above the no results message is way too big at the moment:

| Currently | [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=4524-1318&mode=design&t=cqqqMMwPUdJFi9iN-4) |
| --- | ----------- |
| <img width="1736" alt="Captura de ecrã 2023-10-04, às 16 00 43" src="https://github.com/WordPress/twentytwentyfour/assets/65220155/e1af07b4-d4fe-4b3f-93ef-c4a7f0f06ad5">  |  ![Archive Template_ No results](https://github.com/WordPress/twentytwentyfour/assets/65220155/fcf5584d-260f-47ce-b680-79df96b21570) |

The text also should say instead **"No posts were found."**

beafialho:
Also applies to Search (No results):

<img width="1317" alt="Captura de ecrã 2023-10-04, às 16 27 09" src="https://github.com/WordPress/twentytwentyfour/assets/65220155/b8f9ad26-01b2-4114-a0b3-171a47a09ce0">

See [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=615-10850&mode=design&t=cqqqMMwPUdJFi9iN-4)

luminuu:
Done with #571

PR #571, opened 2023-10-06
Update the 'no results' pattern text and position.

carolinan:
**Description**
For https://github.com/WordPress/twentytwentyfour/issues/566

Updates the text of the no results pattern to "No posts were found".
Moves the pattern higher up in the template, so that it is not pushed down by the post template spacing.

The archive and search uses the same pattern, so I did not try to have different spacing for the two templates.
- In Figma, the search page has 100px spacing between the form and the no results text, and in archives it is 40px between the title and the no results text.
- As a compromise I used preset 30, (2.5 rem 40px) and I added it in theme.json. This can of course be adjusted and moved to the pattern or templates.

**Screenshots**

Search, after:
64 local__s=wolves

Archive, after:
64 local_tag_chattels_ (1)

**Testing Instructions**
View a search and archive page that has no result. 
To view a tag or category with no result, you can go to admin > posts > tags and see a list of tags that has 0 entries. You may need to create one if there are none.

### DIFF
--- a/patterns/hidden-no-results-content.php
+++ b/patterns/hidden-no-results-content.php
@@ -6,5 +6,5 @@
  */
 ?>
 <!-- wp:paragraph -->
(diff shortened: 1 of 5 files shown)


luminuu (approved):
LGTM

Issue #572, opened 2023-10-06
Consider adding a system/web safe font as an alternative to default fonts


LittleBigThing:
Currently, the theme is only using web (Google) fonts and it loads three font files if the theme’s default style is used (400kb!?).

Most other default themes have been using system fonts. IMHO, a default theme should provide a lightweight, less opinionated alternative for users who want to avoid the additional ‘weight’ for their sites. This might have been (kind of/maybe?) possible using the Font Library, but this will not make it into 6.4. 

It would be a nice and easy solution to at least make a system (or a web safe) font available in the typography options. This is simply achieved by registering it in `theme.json`. It could be a [modern font stack](https://modernfontstacks.com) as well, as was suggested in #175, which provides a nice alternative font stack that looks very similar in each operating system (not like system fonts).

luminuu:
Can I get some Design feedback on this @beafialho @richtabor?

MaggieCabrera:
TT4 is not the first default theme to add external fonts to the theme. TT3 was a good example of how to build a lean block theme that looked good without adding much to it and was a nice base for users to build on top of. TT2 did add fonts to the theme files. I think both approaches are okay, and they both show different ways of building themes. They also serve a purpose in how they teach others to replicate the same behavior if they want to add fonts to their themes.

Unless there's a design reason to change the fonts at this stage in development, I don't see a technical reason to make that change so close to RC.

LittleBigThing:
Thanks @MaggieCabrera.

I agree with you about default themes being an example for new WP features and best practices.

I also think that they are examples of giving users freedom of creating. I did not suggest to change the design in this phase of the process, but to simply make a system font available (see PR #573) to give users the freedom of using a ‘weightless’ font if they wish.

I, for example, often use default themes to build lightweight websites. I appreciate having the freedom of not to have to use a Google font for such a site.

I realise that making fonts available is probably the (future) purpose of the Font Library. So that might be a better place to advocate having system/web safe fonts available as alternatives.

MaggieCabrera:
Ok, I misunderstood, and thought we were talking about replacing one of the existing fonts. I think that should be fine, let me comment on that PR

LittleBigThing:
With regard to the comments in the PRs:

Totally agree concerning the confusing name of the fonts. System font and/or System Serif would be much better.

I am not sure either if adding additional (system) fonts or font stacks is theme territory.
[...]

Issue #175, opened 2023-09-01
Consider using a modern font stack instead of system fonts


LittleBigThing:
Several default themes are using system fonts. This is a great practice since they have great language support and it avoids loading several font files. A disadvantage is that they are often not very similar on different operating systems (OSs).

An interesting alternative would be to use a modern font stack. It has the same advantages as a system font stack, but in addition they are chosen to **match better between different OSs**.

The [modern font stacks website](https://modernfontstacks.com) showcases several ideas. For example, the Neo-Grotesque stack could be a great alternative to where the theme now uses system fonts. The GitHub repo of the project demonstrates [how this stack looks](https://github.com/system-fonts/modern-font-stacks#neo-grotesque) in different OSs.

beafialho:
Thank you for the recommendation! I think the Neo-Grotesque stack would be a good choice for paragraphs.

MaggieCabrera:
Let's close this since we swapped to using Inter instead

PR #573, opened 2023-10-06
add system font

LittleBigThing:
Adding a system font as an alternative for web fonts, example solution for #572.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -236,6 +236,11 @@
 					"fontFamily": "Cardo",
 					"name": "Cardo",
 					"slug": "heading"
+				},
+				{
+					"fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
+					"name": "System Font",
+					"slug": "body-alt"
 				}
 			],
 			"fontSizes": [
--- a/theme.json
+++ b/theme.json
@@ -240,7 +240,7 @@
 				{
 					"fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
 					"name": "System Font",
-					"slug": "body-alt"
+					"slug": "system-font"
 				}
 			],
 			"fontSizes": [


theme.json
```diff
@@ -236,6 +236,11 @@
 					"fontFamily": "Cardo",
 					"name": "Cardo",
 					"slug": "heading"
+				},
+				{
+					"fontFamily": "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif",
+					"name": "System Font",
+					"slug": "body-alt"
```
MaggieCabrera: we were using the slugs body/heading so patterns using specific fonts wouldn't have a slug with the name of the font, since that doesn't make sense when you change variations. In this case, since this is a font for end users and not for theme developers, we can keep the slug aligned with the name of the font

MaggieCabrera:
I think you mentioned this idea in https://github.com/WordPress/twentytwentyfour/issues/175 and in the context of the design this theme has, I suggest we include two stacks instead of just one. Let's include a serif and a sans-serif one in case people want to play with those.

What do you think @beafialho, @richtabor ?

LittleBigThing:
I think that would be a great addition to the theme. :-)

I was only thinking about if adding system fonts or font stacks could clash with the Font Library in any way in the future. See a discussion starting [here](https://github.com/WordPress/gutenberg/issues/54186). So I just mention it here. :-)

MaggieCabrera:
> I think that would be a great addition to the theme. :-)
> 
> I was only thinking about if adding system fonts or font stacks could clash with the Font Library in any way in the future. See a discussion starting [here](https://github.com/WordPress/gutenberg/issues/54186). So I just mention it here. :-)

I don't think it will be, most of the additions and features made to the editor are made with backwards compatibility in mind, so they won't implement something that breaks every other block theme that comes before the feature.

@matiasbenedetto can you confirm that this won't be the case?

matiasbenedetto:
> I don't think it will be, most of the additions and features made to the editor are made with backwards compatibility in mind, so they won't implement something that breaks every other block theme that comes before the feature.
> @matiasbenedetto can you confirm that this won't be the case?

Exactly, there won't be any compatibility issues. Themes can add as many fonts as desired. Font Library doesn't change that.

theme.json
LittleBigThing: Changed slug to 'system-font'

LittleBigThing:
See #590 and #591 as well as examples. :-)

MaggieCabrera:
Closed in favor of https://github.com/WordPress/twentytwentyfour/pull/667, thanks for starting this one!

PR #574, opened 2023-10-06
Tweaks for milky style variation

richtabor:
**Description**
Renaming to "Fossil", color tweaks for improved contrast, more unique button styles, fixed header/body fonts missing (we can redeclare the font instead of reassigning header/body fonts. Heading fonts are also set to normal weight, so that they match the bigger paragraph text we're using throughout patterns.

**Screenshots**
CleanShot 2023-10-06 at 14 37 13 CleanShot 2023-10-06 at 14 37 25


luminuu (approved):
LGTM

PR #580, opened 2023-10-07
Hidden patterns should use false instead of no for the inserter property

### LINKED ISSUE #578: Hidden patterns should use `false` instead of `no` for the inserter property
**Description**

For now, hidden pattern use the following declaration to prevent them to be found in the inserter: `Inserter: no`

**Expected behavior**

WordPress default themes should follow the official documentation

> inserter (optional): By default, all patterns will appear in the inserter. To hide a pattern so that it can only be inserted programmatically, set the inserter to false.
source (optional): A string that denotes the source of the pattern. 

https://developer.wordpress.org/block-editor/reference-guides/block-api/block-patterns/

benoitchantre:
**Description**
Fixes  #578.

The official documentation says that `false` should be used instead of `no` for the `inserter` property.

> inserter (optional): By default, all patterns will appear in the inserter. To hide a pattern so that it can only be inserted programmatically, set the inserter to false.
source (optional): A string that denotes the source of the pattern. 

https://developer.wordpress.org/block-editor/reference-guides/block-api/block-patterns/

**Testing Instructions**
Check that hidden patterns are not available from the inserter.

**Contributors**
benoitchantre

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ We do this for patterns we don't want the user to access via the inserter or the
 
 We do this by adding the following line:
 
-` * Inserter: no`
+` * Inserter: false`
 
 Let's prefix hidden patterns using `hidden-` when we name the pattern file.
 
--- a/patterns/hidden-404.php
+++ b/patterns/hidden-404.php
@@ -2,7 +2,7 @@
 /**
  * Title: A 404 page
  * Slug: twentytwentyfour/404
- * Inserter: no
+ * Inserter: false
  */
 ?>
 
--- a/patterns/hidden-comments.php
+++ b/patterns/hidden-comments.php
@@ -2,7 +2,7 @@
 /**
  * Title: Comments
  * slug: twentytwentyfour/comments
- * inserter: no
+ * Inserter: false
  */
 ?>
 
--- a/patterns/hidden-full-screen-image.php
+++ b/patterns/hidden-full-screen-image.php
@@ -2,7 +2,7 @@
 /**
  * Title: Full Screen Image
  * Slug: twentytwentyfour/full-screen-image
- * Inserter: no
+ * Inserter: false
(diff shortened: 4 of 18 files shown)


huzaifaalmesbah:
Thank you for your valuable contribution.

carolinan:
Hi,
the documentation you are refererring to is for `register_block_pattern()` only.
When we place patterns inside the patterns folder, `register_block_pattern()` is not used, see https://developer.wordpress.org/reference/functions/_register_theme_block_patterns/
https://make.wordpress.org/core/2022/05/02/new-features-for-working-with-patterns-and-themes-in-wordpress-6-0/

_It would absolutely have been easier for the developers if the two methods were more consistent_, but now it has already been documented as yes/no and not true/false and it will need to continue to support yes/no.

----
It seems something has broken in 6.4. I would like to learn if this is only something on my local install, or if it can be reproduced, because then we have to create a new issue in the Gutenberg repository.

With 6.4 beta 2, current nightly:

If I add `Inserter: false` OR `Inserter: no` to any pattern in the patterns folder **that is not already hidden**, like if I update patterns/hero.php in my code editor, save the file, and then create a new post or page, that pattern **remains available** to me in the inserter.
Similarly, if I update the categories value for the hero pattern, the category does not change in the editor.

benoitchantre:
Thank you @carolinan for the explanation.

> It seems something has broken in 6.4. I would like to learn if this is only something on my local install, or if it can be reproduced, because then we have to create a new issue in the Gutenberg repository.

Are you using the `WP_DEVELOPMENT_MODE` constant with `theme` as value?
https://make.wordpress.org/core/2023/07/14/configuring-development-mode-in-6-3/

I can't reproduce that when `WP_DEVELOPMENT_MODE` is used.

PR #581, opened 2023-10-07
Fixing Template Part block issue with older versions of WordPress

### LINKED ISSUE #579: WP6.3: No header and footer displayed
**Description**

First, as I can see the theme should be tested with Gutenberg enabled, I'm not even sure if this is a bug.

When I test the theme with WP6.3 _(without Gutenberg plugin active)_, there are no site header and footer template parts displayed in templates that display `patterns\template-*` patterns.

To resolve I have to add `"theme":"twentytwentyfour"` to affected template part blocks in those patterns to make them work.

Is this intentional? Will the theme be incompatible with older WP versions?

webmandesign:
**Description**

Fixes #579

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -178,7 +178,8 @@ if (
 	/**
 	 * Inject the theme attribute into Template Part if needed.
 	 *
-	 * @return void
+	 * @param array $block The block being rendered.
+	 * @return array
 	 * @since Twenty Twenty-Four 1.0
 	 * @see _inject_theme_attribute_in_block_template_content()
 	 * @see gutenberg_render_block_core_pattern()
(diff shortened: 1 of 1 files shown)


carolinan:
The theme is not intended for versions lower than 6.4 because it uses features that are added in 6.4.
If the theme can be activated on 6.3 then something is wrong with how WordPress reads the `Requires at least: 6.4` file header.

luminuu:
Closing this PR, as Carolina said it's not intended to work with 6.3 and lower. I just downloaded it to a 6.3 installation, and I can't activate it there:
CleanShot 2023-10-09 at 10 57 42@2x

webmandesign:
Well, I must have had it active for longer time when the min version was not set to WP6.4. Testing now I can not activate it either.

Issue #537, opened 2023-10-02
Fix all the previous / next navigation and pagination links


afercia:
**Description**

Previous default themes established best practices for all the 'previous / next' navigation links, typically post navigation, paginated post navigation, comments navigation. I'm not sure I see those best practices in place in Twenty Twenty-Four and I'm wondering why.

- All navigation should be wrapped in a properly labeled `nav` element.
- When possible, the `nav` element should also contain a `h2` heading to identify the navigation
- The links text should be meaningful also when read out of context. Using only `Previous` and `Next` is not sufficient.

Some markup examples from Twenty Twenty-One:

Posts navigation:

```
<nav class="navigation post-navigation" aria-label="Posts">
    <h2 class="screen-reader-text">Post navigation</h2>
```

Comments navigation: 

```
<nav class="navigation comments-pagination" aria-label="Comments">
	<h2 class="screen-reader-text">Comments navigation</h2>
```

Paginated post:

```
<nav class="page-links" aria-label="Page"> 
```

Instead, in Twenty Twenty-Four:

1. The navigation links are just `a` elements within `div` elements. No proper semantic and labeling to identify these navigational tools. This is a regression compare to previous default themes.
2. There is no heading to identify these navigation sections.
3. Posts navigation: the links text is only `Previous` and `Next`. There is no sufficient contextual information to understand what these links are about. This is a regression compare to previous default themes.
4. Posts navigation: the links are places _after_ the comments list and comments form. Not sure that is ideal: when there are dozens and dozeens of comments users would need to scroll a lot just fo be able to use the previous / next links.
5. I'm not sure I see any links rendered for the paginated post.

Additionally: I'm not sure I see actual pagination links, as in: Page 2, Page 3, Page 4 etc. I only see Previous/Next, Older Comments/Newer Comments.. I'm not sure whether this is a design choice or a block limitation but I'd think actual pagination linnk would improve the user experience especially when there are many posts and comments.

**Step-by-step reproduction instructions**

- Check the posts navigation links markup and compare with Twenty Twenty-One
- Check the comments pagination links markup and compare with Twenty Twenty-One
- Check the paginated post links markup (in the themes test data is usualy at the URL `2012/01/template-paginated/`) and compare with Twenty Twenty-One

**Expected behavior**

All the navigation and pagination links markup to follow the besst practices established across the years in the previous default themes.

**Screenshots**

Pagianted post: I can't see any pagination links, thus the post content is not accessible:

[...]

PR #583, opened 2023-10-09
Place post navigation in a <nav> and update the link label

carolinan:
**Description**

Partial for https://github.com/WordPress/twentytwentyfour/issues/537

Updates the post navigation link (next and previous post):
- Moves the markup to a pattern
- Changes the wrapping div to a nav
- Includes the post title as part of the link text

**Screenshots**
Updated next and previous post labels 2023-10-09

**Testing Instructions**
Make sure your WordPress installation has enough content so that you can navigate to both the next and previous posts.
Confirm that both the "next" and "previous" labels as well as the post title title and arrow shows.

- View a single post on the front and the template in the Site Editor
- View a **single post with a sidebar** on the front and the template in the Site Editor

### DIFF
--- a/patterns/hidden-post-navigation.php
+++ b/patterns/hidden-post-navigation.php
@@ -8,7 +8,7 @@
 
 <!-- wp:group {"tagName":"nav","ariaLabel":"<?php esc_attr_e( 'Posts', 'twentytwentyfour') ?>","style":{"spacing":{"padding":{"bottom":"0vh"}}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
 <nav class="wp-block-group" style="padding-bottom:0vh" aria-label="<?php esc_attr_e( 'Posts', 'twentytwentyfour') ?>">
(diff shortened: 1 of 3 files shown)


MaggieCabrera:
I'm fine with doing this on the theme for now but this is an issue in the block, it should have the correct markup by default, same as the nav block uses the right tag for it. Is there an open issue in GB for this?

carolinan:
I could not find an issue for it.

carolinan:
In case it is not obvious, the proposed design is for the custom text "Next: " to not be linked. It is linked and underlined here because there is a bug with the block.

The custom text and the post title are both inside the `<a>`, element, in their own `<span>` with a class name. But I did not attempt to hide the underline of the custom text because in my **opinion** it would be difficult to create a good looking style for both the default state and the :hover state.

patterns/hidden-post-navigation.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Title: Post navigation
+ * slug: twentytwentyfour/post-navigation
```
MaggieCabrera, line 4: should this be a hidden one?

luminuu: The other hidden patterns look the same, it says `hidden` in the file name and `inserter: no` below, so it's looking good to me?

MaggieCabrera: forgive my blindness and forget my comment :D

luminuu (approved):
LGTM

PR #586, opened 2023-10-09
Tweak spacing presets on mobile

MaggieCabrera:
**Description**

Fixes the issue commented [here](https://github.com/WordPress/twentytwentyfour/pull/557#issuecomment-1746394168).

This PR makes the preset 10 to always be 1rem. This should keep the desktop version looking the same but improve the mobile version. A lot of patterns and templates use this preset and I generally found that this was an improvement to what we had before.

**Screenshots**

Before | After 
--- | ---
Screenshot 2023-10-09 at 18 25 27 | Screenshot 2023-10-09 at 18 26 54

Before | After 
--- | ---
Screenshot 2023-10-09 at 18 45 52 | Screenshot 2023-10-09 at 18 31 48

Before | After 
--- | ---
Screenshot 2023-10-09 at 18 45 40 | Screenshot 2023-10-09 at 18 34 24

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -162,7 +162,7 @@
 			"spacingSizes": [
 				{
 					"name": "1",
-					"size": "min(1rem, 1vw)",
+					"size": "1rem",
 					"slug": "10"
 				},
 				{


luminuu (approved):
LGTM

PR #590, opened 2023-10-10
add an Old Style font stack

LittleBigThing:
Add an Old Style (serif) font stack (based on modernfontstacks.com) as a light-weight alternative for Cardo

Example implementation of #572

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -236,6 +236,11 @@
 					"fontFamily": "Cardo",
 					"name": "Cardo",
 					"slug": "heading"
+				},
+				{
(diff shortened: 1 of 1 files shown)


richtabor:
I'm not sure if we should include font stacks within the default variation alone; but rather alongside alternate style variations.

richtabor:
I'm not 100% on this, but if we do we should probably name it something more recognizable, like "System Serif" or something.

luminuu:
I'd prefer it if fonts are used that are actually available by default on a wide range of operating systems, for example I have none of these installed on my Mac. If that is not possible, I'd prefer not to bring this in.

MaggieCabrera:
Closed in favor of https://github.com/WordPress/twentytwentyfour/pull/667

PR #591, opened 2023-10-10
add a Neo-Grotesque font stack

LittleBigThing:
Add a Neo-Grotesque (sans-serif) font stack (based on modernfontstacks.com) as a light-weight alternative for Inter

Example implementation of #572

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -236,6 +236,11 @@
 					"fontFamily": "Cardo",
 					"name": "Cardo",
 					"slug": "heading"
+				},
+				{
+					"fontFamily": "Inter, Roboto, 'Helvetica Neue', 'Arial Nova', 'Nimbus Sans', Arial, sans-serif",
+					"name": "Neo-Grotesque",
+					"slug": "neo-grotesque"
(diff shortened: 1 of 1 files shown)


richtabor:
We should probably name this something more recognizable, like "System font".

luminuu:
There is a PR for adding the true system font: #573 - I'd prefer that one over this PR here, so we don't have too many system fonts cluttering the interface.

richtabor:
Definitely only one. But if we add it to the parent, we'll need it on any variation that overrides fonts I think. Should double-check.

luminuu:
@richtabor Just tested this, we only need to udpate the style variations adding the system fonts if they have fonts declared in the style variation json. If there's no explicit declaration, the fonts from the main `theme.json` are used. Which means Fossil, Ice, Mint and Sandstorm need to be updated. Onyx and Rust need no work.

MaggieCabrera:
Closed in favor of https://github.com/WordPress/twentytwentyfour/pull/667

PR #594, opened 2023-10-10
Improving search form wrapper

### LINKED ISSUE #593: Improve search results page search form container

In `templates/search.html` and `patterns/template-search-portfolio.php` the search form is placed within Columns block.
The first column is 66.66% wide and contains the search form.
The second column is 33.33% and contains Spacer block with no height. (Why is there this Spacer block?)

We can possibly simplify the HTML here.

Instead of using Columns block as wrapper for search block, we could use Group block like so:
```html
<!-- wp:group {"align":"wide","layout":{"type":"constrained","contentSize":"840px","justifyContent":"left"}} -->
<div class="wp-block-group alignwide">
	<!-- wp:pattern {"slug":"twentytwentyfour/search"} /-->
</div>
<!-- /wp:group -->
```

webmandesign:
**Description**

Fixes #593

Tested on WP 6.4-beta2 with Gutenberg 16.7.1.

### DIFF
--- a/patterns/template-search-portfolio.php
+++ b/patterns/template-search-portfolio.php
@@ -10,28 +10,15 @@
 
 <!-- wp:template-part {"slug":"header"} /-->
 
-
 <!-- wp:group {"tagName":"main","align":"full","layout":{"type":"constrained"}} -->
 <main class="wp-block-group alignfull">
 	<!-- wp:query-title {"type":"search","align":"wide","style":{"typography":{"lineHeight":"1"},"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|10"}}}} /-->
 
-	<!-- wp:columns {"align":"wide"} -->
-	<div class="wp-block-columns alignwide">
-		<!-- wp:column {"width":"66.66%"} -->
-		<div class="wp-block-column" style="flex-basis:66.66%">
-			<!-- wp:pattern {"slug":"twentytwentyfour/search"} /-->
-		</div>
-		<!-- /wp:column -->
-
-		<!-- wp:column {"width":"33.33%"} -->
-		<div class="wp-block-column" style="flex-basis:33.33%">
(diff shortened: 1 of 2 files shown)


patterns/template-search-portfolio.php
```diff
@@ -10,28 +10,15 @@
...
-		<!-- /wp:column -->
-
-		<!-- wp:column {"width":"33.33%"} -->
-		<div class="wp-block-column" style="flex-basis:33.33%">
-			<!-- wp:spacer {"height":"0"} -->
-			<div style="height:0" aria-hidden="true" class="wp-block-spacer"></div>
-			<!-- /wp:spacer -->
-		</div>
-		<!-- /wp:column -->
+	<!-- wp:group {"align":"wide","layout":{"type":"constrained","contentSize":"840px","justifyContent":"left"}} -->
```
MaggieCabrera, line 17: I'm not a fan of the magic number but it's definitely better than the empty spacer in a column. Looking in the editor I think we can remove the columns altogether and apply the width to the search block. We won't be able to use the pattern, but this is already a pattern so the text will be translated too. Let's do that and use the percentage instead of the static px value, what do you think?

webmandesign: Percentage can not be used here due to responsive design. The percentage value would be applied on small/narrow screens too. That's why more precise value is better here - it works as `max-width`.

webmandesign: I've tested the layout you've proposed, but it won't work.

The wrapper Group block here is not only to set the max-width for the search form, but also to make the search form align with wide aligned H1 (query title) above.

The search form block does not have wide alignment option either.

MaggieCabrera: yeah, I tested yours and it works better, so let's go with this! since this is only happening here we can work with having "one" magic number 😆

MaggieCabrera (approved):

PR #601, opened 2023-10-10
Re-add area and tagName attributes to header template part calls

### LINKED ISSUE #600: Re add area to header template parts
When we removed the sticky from the header we removed the `"area":"header`" by mistake, let's add it back. Also, we probably should add the header tagname, since we don't have it anymore!

luminuu:
`area` and `tagName` attributes are added again to the header template part calls in all templates and patterns. Fixes #600.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,11 +153,11 @@ turns into
 
 and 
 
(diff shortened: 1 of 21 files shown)


MaggieCabrera (approved):
Thank you!

PR #603, opened 2023-10-10
Rename footer template parts and patterns

### LINKED ISSUE #411: Alternate template/template part implementation
Derived from a number of questions and comments throughout the project, one thing that's clear is that we need to iron out our approach to alternate templates (business, portfolio, writer home/blog/single templates). 

## Templates

Currently, TT4 has templates for each "use case" available in the site editor entirely. 

You can apply a different template to a _specific_ page within the post editor, and with Gutenberg 16.7, you can [swap templates](https://github.com/WordPress/gutenberg/pull/51477) for other like templates (such as page templates). Swapping from template, to template, is not yet available (Blog Home -> Alternate Blog Home), but it's not a big ask to add the existing functionality to the Template Inspector https://github.com/WordPress/gutenberg/issues/54562. 

To enable template swapping, templates need to be abstracted into patterns, [like in this pr](https://github.com/WordPress/twentytwentyfour/pull/322/files). We can have patterns that are essentially template content ([page-01-business-home.php](https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/page-01-business-home.php)), and corresponding templates that reference the pattern ([home.html](https://github.com/WordPress/twentytwentyfour/blob/trunk/templates/home.html)). 

This is fine for the _one_ home, or archive page, but if we have each of the alternative archive designs (writer and portfolio) as individual templates, we start to see where this model is fragile. **Why expose two archive pages that are irrelevant to the site as templates in the site editor?** 

You can't use the "Archive Portfolio" template instead of "Archive Writer" — not without https://github.com/WordPress/gutenberg/issues/54562. But when that issue is resolved, it pushes forward the notion of swapping, rather than providing those all top-level as templates. 

**I propose we make all alternative index/archive/search designs as patterns in the template swap flow.** And for single/page designs, either rename to make them more generic, or do the same. 

### Visuals 

Current site editor templates view. There are more templates in the queue to add even:
CleanShot 2023-09-18 at 14 41 21

Selecting an alternate template within the post editor:
CleanShot 2023-09-18 at 14 42 48

Selecting an alternate template within the site editor:
CleanShot 2023-09-18 at 14 46 07

## Naming templates and template parts

However we decide to move forward with the template implementation, we should consider renaming templates, patterns, and template parts to reduce the visibility of "Business", "Portfolio", and "Writer" terminology. This would follow suite with how core patterns are named as well, where the actual title of the pattern helps identify what it is. And also allow for more flexible use cases. 

Examples (drafts): 
Footer portfolio (pattern) -> Three column footer with colophon
Footer writer (pattern) -> Minimal centered footer with logo and navigation
Page portfolio (template) -> Wide image with title content below

And perhaps this helps clear the confusion with the lot of templates provided. 

## Template parts

TT4 also has multiple header and footer designs, originally intended to be applied to corresponding templates. I.e. the "Footer Portfolio" design would be applied to the "Home Portfolio" template. 

I suggest we don't do this, but rather only provide one "footer" template part (and header), which people then have the ability to replace the footer with the provided alternatives. Using this method, the header/footer parts on a site are always in sync — unless the user _chooses_ to create and display a different header on a page. 

When you change a footer to use the "Footer Portfolio" design, it will be applied throughout the site, rather than having one template with "Footer Portfolio" and another with "Footer Writer". 

## Related: 

https://github.com/WordPress/twentytwentyfour/pull/406#issuecomment-1724152852, https://github.com/WordPress/twentytwentyfour/issues/9#issuecomment-1724150856, https://github.com/WordPress/twentytwentyfour/issues/351#issuecomment-1719057371, https://github.com/WordPress/twentytwentyfour/pull/368#issuecomment-172252390

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/411

This renames the patterns to be descriptive in the UI:

Screenshot 2023-10-10 at 17 58 48

### DIFF
--- a/patterns/footer.php
+++ b/patterns/footer.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Default footer
+ * Title: Three column footer with colophon
  * Slug: twentytwentyfour/footer
  * Categories: footer
  * Block Types: core/template-part/footer
--- a/parts/footer-portfolio.html
+++ b/parts/footer-portfolio.html
@@ -1,1 +1,0 @@
-<!-- wp:pattern {"slug":"twentytwentyfour/footer-portfolio"} /-->
--- a/parts/footer-writer.html
+++ b/parts/footer-writer.html
@@ -1,1 +1,0 @@
-<!-- wp:pattern {"slug":"twentytwentyfour/footer-writer"} /-->
--- a/patterns/footer-portfolio.php
+++ b/patterns/footer-portfolio.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Footer portfolio
+ * Title: Two column footer with colophon
  * Slug: twentytwentyfour/footer-portfolio
  * Categories: footer
  * Block Types: core/template-part/footer
--- a/patterns/footer-writer.php
+++ b/patterns/footer-writer.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Footer writer
+ * Title: Minimal centered footer with logo and navigation
  * Slug: twentytwentyfour/footer-writer
  * Categories: footer
  * Block Types: core/template-part/footer
--- a/theme.json
+++ b/theme.json
@@ -922,17 +922,7 @@
 		{
 			"area": "footer",
 			"name": "footer",
-			"title": "Three column footer with colophon"
-		},
-		{
-			"area": "footer",
-			"name": "footer-portfolio",
-			"title": "Two column footer with colophon"
-		},
-		{
-			"area": "footer",
-			"name": "footer-writer",
-			"title": "Minimal centered footer with logo and navigation"
+			"title": "Footer"
 		},
 		{
 			"area": "uncategorized",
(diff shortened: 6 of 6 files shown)


patterns/footer.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Default footer
+ * Title: Three column footer with colophon
```
luminuu, line 3: Should we put "Default:" as a prefix here, so users can easily spot which footer is set by default to the theme?

carolinan: When I read these names it makes me think that the editor needs to be updated to show the description by default. It is visually hidden now.

kafleg: Yes, it is always confusing with the name.

MaggieCabrera:
I talked to @richtabor about this and right now when we replace the content of the footer we have them all duplicated because we have template parts and patterns, so I removed the former.

patterns/footer.php
MaggieCabrera: I think we don't need the word, since when we clear customizations we will be back to that one. In the end the important bit is that the user finds one that they like, it doesn't matter if it is the default one or not.

luminuu (approved):
LGTM

PR #604, opened 2023-10-10
Adjust single with sidebar template

### LINKED ISSUE #599: Search block on sidebar
The search block looks like it's extending over the width of the sidebar and the button doesn't have border radius. Let's make sure we are styling the button element to look the same as the button block.

Screenshot 2023-10-10 at 16 41 05

It should look like this:

image

luminuu:
**Description**

I've updated the Single with Sidebar template to get the sidebar close to 360 pixels as in the design. It's not exact, but removing the empty columns and adding some extra padding brings it closer to the design. The sidebar is about 340 pixels at max.

**Screenshots**

CleanShot 2023-10-10 at 19 23 28@2x

**Testing Instructions**

1. Checkout this PR
2. Go to the Site Editor -> Templates
3. Make sure there are no customizations saved for the "Single with Sidebar" template
4. Open template to check in the editor
5. Open a post
6. Set the template to "Single with Sidebar"
7. Save post
8. Open saved post in the frontend


templates/single-with-sidebar.html
```diff
@@ -1,60 +1,47 @@
-<!-- wp:template-part {"slug":"header","area":"header","tagName":"header"} /-->
+<!-- wp:template-part {"slug":"header","theme":"twentytwentyfour","tagName":"header","area":"header"} /-->
```
MaggieCabrera: the theme slug should be gone

luminuu: Ah, that must have been some copy/pasting error. Done.

MaggieCabrera (approved):
I fixed an issue where the post title wouldn't be aligned with the featured image and the post content. Thanks for doing this, now the markup is much simpler and it looks closer to the design

PR #606, opened 2023-10-11
Updated styles for plain quote"

### LINKED ISSUE #439: Quote (Plain): Improve appearance
Currently the Plain Quote is looking a bit awkward. Let's improve its appearance by setting it to use `System font Semi Bold` only for the quote, citation should remain `Regular`. The line height should be `1,6`. Alternatively, see [Figma](https://www.figma.com/file/AlYr03vh4dVimwYwQkTdf6/Twenty-Twenty-Four?type=design&node-id=565-9983&mode=design&t=mlxcmncwpkBhDVAw-4).

| Currently      | Intended |
| ----------- | ----------- |
|  Captura de ecrã 2023-09-20, às 12 04 18    |   Captura de ecrã 2023-09-20, às 12 06 32   |

This one seems to look the best default for the quote in a paragraph. If we want to just create a post that's a quote though, like the example below, is it possible to make the larger one default for such cases?

Captura de ecrã 2023-09-20, às 12 22 40

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/439

I tried to maintain the CSS to a minimum and not break RTL

**Screenshots**

RTL | LTR
--- | ---
 Screenshot 2023-10-11 at 11 28 50 | Screenshot 2023-10-11 at 11 30 22

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -660,7 +660,7 @@
 				"color": {
 					"background": "var(--wp--preset--color--base-2)"
 				},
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
In the end, the amount of CSS wasn't too bad, I think we can bring this one in.

luminuu (approved):
LGTM

PR #608, opened 2023-10-11
Removing obsolete Group block wrappers

### LINKED ISSUE #587: Remove obsolete Group block wrappers

There is an obsolete Group block wrapper applied on `post-title` block in `single.html` and `single-with-sidebar.html` template.

We can remove this Group block as it does not provide any styling or layout benefits.

webmandesign:
**Description**

Fixes #587

Tested with WP 6.4-beta2 with Gutenberg 16.7.1.

### DIFF
--- a/templates/single.html
+++ b/templates/single.html
@@ -48,6 +48,8 @@
 			<!-- /wp:group -->
 		</div>
 		<!-- /wp:group -->
+	</div>
+	<!-- /wp:group -->
 </main>
 <!-- /wp:group -->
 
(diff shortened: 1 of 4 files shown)


templates/single.html
```diff
@@ -55,4 +51,4 @@
 </main>
```
luminuu, line 53: Before the closing `main`, there's a closing group tag/comment missing:

```
</div>
<!-- /wp:group -->
```

luminuu (changes requested):

templates/single.html
webmandesign: Interesting. This must have been missing for some time. I've added the closing Group tag here.

webmandesign: Great catch!

luminuu (approved):
LGTM

PR #610, opened 2023-10-11
Fix global heading sizes

### LINKED ISSUE #551: H5 and H6 is not having any difference.
**Description**

If you check in Heading block with H4 to H6 there is no difference visible between both.

1. Take Two Heading block for comparison.
2. For first one choose h5 and second one choose h6
3. Use same text so see difference clearly

**Expected behavior**

H5 and H6 is having a little bit of change occur when we change the heading level either font-weight or font-size.

**Video**

Video URL :- https://share.cleanshot.com/Mr7ZkwNfBXprgNqgxwl1

**Environment info**

- Mac OS
- Chrome Browser
- Wordpress 6.3.1
- Gutenberg - 16.7.0

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/551

This PR adds replicates the flexible font size for the h4 heading so it scales like the rest of the headings.

H4 and H5 are used on two patterns but both of them use a preset font size instead of the global value, so they are unaffected by this change. H6 is never used on templates or patterns.

**Screenshots**

**Desktop**

Screenshot 2023-10-11 at 13 20 40

**Mobile**

Screenshot 2023-10-11 at 13 20 17

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -867,7 +867,7 @@
 			},
 			"h4": {
 				"typography": {
-					"fontSize": "var(--wp--preset--font-size--medium)"
+					"fontSize": "clamp(1.1rem, 1.1rem + ((1vw - 0.2rem) * 0.767), 1.5rem)"
 				}
 			},
 			"h5": {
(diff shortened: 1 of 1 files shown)


luminuu (approved):
LGTM

PR #614, opened 2023-10-11
Fix regressions on single template

### LINKED ISSUE #612: Post template visual improvements
I don't think the post template was looking like this previously, but if we look at the Figma designs, it currently has a lot of inconsistencies:

- Tags block appears twice
- The separator has the wrong color and it should be wide width
- The "Comments" heading should not be there
- There's a lack of padding in the "Leave a reply" section and in the post pagination

Here's a recording of what I mean and what should be adjusted:

https://github.com/WordPress/twentytwentyfour/assets/65220155/464ebb75-bf7c-4d76-ba68-6c6ad7801693

MaggieCabrera:
**Description**

Closes https://github.com/WordPress/twentytwentyfour/issues/612

Some of these regressions come from bad conflict-solving from the formatting refactor. 

**Screenshots**

Screenshot 2023-10-11 at 17-35-56 Molestiae aut ab labore – wp-stable

### DIFF
--- a/patterns/hidden-comments.php
+++ b/patterns/hidden-comments.php
@@ -44,6 +44,6 @@
 	<!-- wp:comments-pagination-next /-->
 	<!-- /wp:comments-pagination -->
 
-	<!-- wp:post-comments-form /-->
+	<!-- wp:post-comments-form {"style":{"spacing":{"padding":{"top":"var:preset|spacing|20","bottom":"var:preset|spacing|20"}}}} /-->
 </div>
 <!-- /wp:comments -->
--- a/patterns/hidden-post-navigation.php
+++ b/patterns/hidden-post-navigation.php
@@ -6,9 +6,9 @@
  */
 ?>
 
-<!-- wp:group {"tagName":"nav","ariaLabel":"<?php esc_attr_e('Posts', 'twentytwentyfour'); ?>","style":{"spacing":{"padding":{"bottom":"var:preset|spacing|40","top":"var:preset|spacing|40"}}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
-<nav class="wp-block-group" style="padding-top:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--40)" aria-label="<?php esc_attr_e('Posts', 'twentytwentyfour'); ?>">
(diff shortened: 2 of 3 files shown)


MaggieCabrera:
@beafialho does this look good to you?

beafialho:
Do we have to have that huge "Comments" heading?

Sorry @MaggieCabrera one more thing. As mentioned in [this comment](https://github.com/WordPress/twentytwentyfour/issues/537#issuecomment-1744693039), the post pagination should ideally show the post title **only in the Portfolio template**. If that's not possible, the label should just say "Next Post" and "Previous Post".

MaggieCabrera:
The comments heading needs to be there for accesibility reasons, and I think so are the titles in the post navigation, right @carolinan?

webmandesign:
Maybe a new block style could be registered to hide the comments heading and possibly even post navigation post title accessibly?

luminuu:
> Maybe a new block style could be registered to hide the comments heading and possibly even post navigation post title accessibly?

Hiding any heading breaks the accessibility and will make the theme not get the `accessibility-ready` tag. We're not changing it.

webmandesign:
> Hiding any heading breaks the accessibility and will make the theme not get the `accessibility-ready` tag. We're not changing it.

I'm suggesting **accessible** hiding here, something like WP native `.screen-reader-text` class:
```css
.screen-reader-text {
	clip: rect(1px, 1px, 1px, 1px);
	word-wrap: normal !important;
	border: 0;
	-webkit-clip-path: inset(50%);
	clip-path: inset(50%);
	height: 1px;
	margin: -1px;
	overflow: hidden;
	padding: 0;
	position: absolute;
	width: 1px
}
```

If I'm not mistaken (and nothing has changed for `accessibility-ready` tag review process), such accessible hiding is allowed in themes. Or am I missing something @carolinan?

luminuu:
No it was this way and was removed, please see this PR: https://github.com/WordPress/twentytwentyfour/pull/500

It also introduces block validation errors done this way.

webmandesign:
@luminuu Sorry, I should explain my suggestion in more details.

I understand using the `screen-reader-text` class directly is not ideal.

I suggest [creating/registering a dedicated block style](https://github.com/WordPress/twentytwentyfour/blob/trunk/functions.php#L23) which would mimic the behavior.

I personally do this in my themes and in [ABS plugin](https://wordpress.org/plugins/additional-block-styles/). Here is the result:

a11y-hidden-block-style

luminuu (approved):
I'm merging this PR now as is. We can't change the links in the post navigation, as these come from a shared pattern across all templates where it's used. As far as I know it's also better to have the post title shown and not just Next/Previous link for accessibility reasons.

PR #630, opened 2023-10-12
Add offset image grid patterns

richtabor:
**Description**

Adds patterns that leverage the neat portfolio grid, but instead of query loop it's image blocks.

Otherwise, you can’t actually make a portfolio _with_ a blog as-is. This would allow you to add a separate portfolio gallery to any page. 

**Screenshots**

CleanShot 2023-10-12 at 15 21 41

### DIFF
--- a/patterns/gallery-offset-image-grid-2-col.php
+++ b/patterns/gallery-offset-image-grid-2-col.php
@@ -1,7 +1,7 @@
 <?php
 /**
- * Title: Offset Image Grid, 3 Columns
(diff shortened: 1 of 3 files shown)


richtabor:
Note: you may need to set `define( 'WP_DEVELOPMENT_MODE', 'theme' );` to see changes quickly due to improved pattern caching in https://github.com/WordPress/WordPress/commit/b0872b005dd27baad7b0c75618baa99a5b10368f

luminuu (approved):
LGTM

PR #637, opened 2023-10-12
Add default block names to a few specific patterns

### LINKED ISSUE #515: Rename blocks in patterns
The rename block feature is going to land on 6.4, so I propose we use it on our patterns so they are clear to the users as to what they are. We can rename the main container of the pattern and in more complex ones any relevant sections (like rename the sidebar column and the main column on the templates that use them as layout!)

My main concern with this is that adding this may produce broken markup for the theme (which already happens for patterns that use new features like aspect ratio) if you are using a previous version of WordPress. What should we do about that?

Screenshot 2023-09-26 at 17 59 49

richtabor:
**Description**

Closes #515 by adding names to some blocks. 

**Screenshots**

CleanShot 2023-10-12 at 16 10 14

**Testing Instructions**

Search for the edited patterns, insert them, see block names.

### DIFF
--- a/patterns/clients.php
+++ b/patterns/clients.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Clients
+ * Title: Client logos
  * Slug: twentytwentyfour/clients
  * Categories: testimonials, about
  * Keywords: logos, social proof, reviews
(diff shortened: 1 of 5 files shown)


luminuu (approved):
LGTM

PR #641, opened 2023-10-13
Tweak quote spacing

### LINKED ISSUE #620: Quote Block without citation has different gaps on the Top and Bottom
Hi,

I noticed that when using the "Quote Block" without adding any citations, the top and bottom spacing doesn't look quite right. It doesn't look very good visually. Can this be fixed?

Thanks,

The issue with the unequal spacing in the "Quote Block" seems to be related to a default margin on the bottom part of the "P" tag. In my testing with TT3, I noticed they've applied a default margin to both the top and bottom, which looks better. Perhaps we could consider adjusting the default gaps for citation text instead of focusing on the "P" tag.
Testing-WP-6-4-and-TT4

enodekciw:
**Description**

When no `<cite>` exists on `core/quote` block, margin-bottom of the last child element is stacking with `<blockquote>` padding-bottom. This should make Quote block spacing more consistent.

Closes #620 

**Before and After**

Original layout:
core-quote-before

Updated layout:
core-quote-after

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -106,3 +106,4 @@
 | @alishabajra | @alishabajracharya27 |
 | @Yos0107 | @yosephtamang |
 | @gregoryfu |  |
+| @enodekciw | @enodekciw |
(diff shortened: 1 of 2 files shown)


kafleg:
Thank you for creating this PR. It is working as expected. 

Editor view:
image


Front view:
image

MaggieCabrera (approved):

PR #642, opened 2023-10-13
Rename Writer patterns/templates to Blogging

### LINKED ISSUE #624: Change references to "Writer" template
I propose we simplify the naming convention a bit for the writer template. The others are classification of sites, i.e. a business site, or a portfolio site, whereas "writer" is not a type of site.

Perhaps we should rename it to "Blog" instead. 

CleanShot 2023-10-12 at 10 53 53

luminuu:
I've renamed all "Writer" references in patterns and templates to "Blogging". I decided against "Blog", as something like "Blog Home", "Blog Search", "Blog Index", "Blog Archive" would be too confusing with already existing WordPress terminology.

### DIFF
--- a/patterns/page-06-blogging-home.php
+++ b/patterns/page-06-blogging-home.php
@@ -1,7 +1,7 @@
 <?php
 /**
- * Title: Writer Home Page
- * Slug: twentytwentyfour/writer-home
+ * Title: Blogging Home
+ * Slug: twentytwentyfour/blogging-home
  * Categories: text, page
  * Keywords: page, starter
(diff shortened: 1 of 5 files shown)


MaggieCabrera (approved):
Ok, I arrived to this one after reviewing the other PR about renaming. I'm fine with this if @richtabor and @beafialho agree too. Let's just keep it consistent.

luminuu:
Just merged the latest trunk, so it can be merged now.

@richtabor Can you have a quick look at this?

richtabor (approved):

Issue #569, opened 2023-10-06
Sandstorm: Focused "outline" button does not have background color


carolinan:
**Description**
This may be a non issue. It is not an accessibility issue but I want to mention it because the styles are inconsistent.

Variation: Sandstorm
Button style: Outline

When you hover over a outlined button, the background and text color changes.
When you focus on the outlined button, the only visible difference is that the outline is added.

In comparison, when you focus on the filled button, it does change background and text color, and the outline is added.

**Step-by-step reproduction instructions**
Add two buttons, with links. One should use the default filled style, the other outline.

**Expected behavior**
For the hover and focus styles to be more consistent(?)
Like I said this might be a non-issue. I want to mention it in case it is not intentional.

carolinan:
I am seeing the same on Onyx, Mint.
Rust has the same inconsistency but also no hover style for the filled button.

luminuu:
@beafialho Can you give some feedback/suggestions on this?

beafialho:
Good catch. I tried to exemplify on Figma how these should look across variations. Is this clear?

<img width="1022" alt="Captura de ecrã 2023-10-11, às 12 05 48" src="https://github.com/WordPress/twentytwentyfour/assets/65220155/01d6bfb0-c131-40f7-a424-41f36768dc5f">

luminuu:
@beafialho Can I assume you want the buttons to always:
- On hover, use `contrast-2` as background color for the filled buttons, and text and outline color for the outlined buttons
- On Focus, use the same as above with the defined outline settings

I'll add the current status with screencasts in a follow up comment, so it will be easier to see where work needs to be done.

luminuu:
**Default, Ice, Mint:**

https://github.com/WordPress/twentytwentyfour/assets/7585600/58ff6c96-de1b-427d-9e2d-c9101c078c61


Uses `contrast` (black) as filled background/outline color and `contrast-2` (grey) as hover state background/outline color. For the outlined button, the focus state is incorrect and needs to be `contrast-2` too.

**Fossil:**

https://github.com/WordPress/twentytwentyfour/assets/7585600/17d61d66-5802-4170-b339-bd5c6cf6d222

Uses `contrast-2` as default color and `contrast` as hover color. Since the `focus` state is not explicitly defined, it uses again `contrast-2` from the default style variation for the focus state. Same for the outlined style variation.

**Onyx:**

[...]

PR #643, opened 2023-10-13
Fix button padding in style variations

luminuu:
As discussed in #569, the size of buttons in the Sandstorm and Fossil style variations should match with all other buttons. Removing extra padding from the style variations.

### DIFF
--- a/styles/fossil.json
+++ b/styles/fossil.json
@@ -165,14 +165,6 @@
 					"background": "var(--wp--preset--color--contrast-2)",
 					"text": "var(--wp--preset--color--white)"
 				},
-				"spacing": {
-					"padding": {
-						"bottom": "0.9rem",
-						"left": "2rem",
-						"right": "2rem",
-						"top": "0.9rem"
-					}
-				},
 				"typography": {
 					"fontFamily": "var(--wp--preset--font-family--heading)",
 					"fontSize": "var(--wp--preset--font-size--small)",
--- a/styles/sandstorm.json
+++ b/styles/sandstorm.json
@@ -127,14 +127,6 @@
 						}
 					}
 				},
-				"spacing": {
-					"padding": {
-						"bottom": "var(--wp--preset--spacing--10)",
-						"left": "var(--wp--preset--spacing--20)",
-						"right": "var(--wp--preset--spacing--20)",
-						"top": "var(--wp--preset--spacing--10)"
-					}
-				},
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--medium)",
 					"textTransform": "capitalize"


richtabor (changes requested):
I don't agree. Buttons can be different sizes in variations. 

> We should have the buttons be the original size like the outline.

The problem is that "outline" variations use core padding styles, not what is defined in the theme.json button styles. So the outline variations are different sizes. 

The problem is not that buttons are different sizes in variations — but that's not something we can fix in the theme (easily). 

I think we need even more variety across variations, and button styles are the easiest way to add that.

richtabor:
It is possible to force our theme.json values on the outline button, but I think it's probably a bit too much lifting out of the box. Here's a rough draft of what that could look like. Then you'd need to enqueue it with wp_add_inline_style perhaps within the editor and the front end. 

I don't _think_ we should do this — what do you think? 

```
	function twentytwentyfour_outline_button_padding() {

		$padding = \WP_Theme_JSON_Resolver::get_merged_data()->get_raw_data()['styles']['elements']['button']['spacing']['padding'];

		$top    = isset( $padding['top'] ) ? "padding-top: {$padding['top']};" : null;
		$right  = isset( $padding['right'] ) ? "padding-right: {$padding['right']};" : null;
		$bottom = isset( $padding['bottom'] ) ? "padding-bottom: {$padding['bottom']};" : null;
		$left   = isset( $padding['left'] ) ? "padding-left: {$padding['left']};" : null;

		$output = '';

		if ( isset( $padding ) ) {

			$output .= '
				.wp-block-button.is-style-outline > .wp-block-button__link {
					' . esc_attr( $top . $right . $bottom . $left ) . '
				}
			';

		}

		return $output;

	}
```

CleanShot 2023-10-13 at 10 01 10

luminuu:
I've just done this PR as per request from @beafialho in the issue.

I personally think we should not add custom functionality via PHP to change something that is defined primarily in theme.json - it will just generate too much confusion for the user and fellow theme developers when they want to change the padding but can't.

Either we go with this PR or we don't. As much as I appreciate the input, with just a weekend short of RC1, I want to get things smoothened out as much as possible.

MaggieCabrera:
So:

- I agree that having different sized buttons between variations is absolutely fine
- I wouldn't add the PHP to fix the size difference with outlined buttons and I'd rather keep said difference if that may be

So I'd rather not go with this one at all. Do we have patterns that have an outlined and a filled button together? I would manually fix those maybe and leave the defaults as they are.

richtabor:
> So:
> 
> 
> 
> - I agree that having different sized buttons between variations is absolutely fine
> 
> - I wouldn't add the PHP to fix the size difference with outlined buttons and I'd rather keep said difference if that may be
> 
> 
> 
> So I'd rather not go with this one at all. Do we have patterns that have an outlined and a filled button together? I would manually fix those maybe and leave the defaults as they are.

Agreed. I think there's one pattern.

MaggieCabrera:
Closed in favor of https://github.com/WordPress/twentytwentyfour/pull/646

PR #644, opened 2023-10-13
Removing "Template" from template file title and add "Business" to page templates

### LINKED ISSUE #623: Remove duplicative "template" title from template patterns
I don't think it's necessary to include the "template" label appended to each template. The UI already indicates that these are templates. 

CleanShot 2023-10-12 at 10 52 38

luminuu:
Renaming pattern/template titles for better consistency.

### DIFF
--- a/patterns/template-archive-writer.php
+++ b/patterns/template-archive-writer.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Writer Archive Template
+ * Title: Writer Archive
  * Slug: twentytwentyfour/template-archive-writer
  * Template Types: archive, category, tag, author, date
  * Viewport width: 1400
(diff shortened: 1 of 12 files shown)


patterns/template-archive-writer.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Writer Archive Template
+ * Title: Writer Archive
```
MaggieCabrera, line 3: We renamed the writer one for another pattern, are we keeping the name for these ones?

MaggieCabrera (approved):
I made a comment but it's not a blocker for me

patterns/template-archive-writer.php
luminuu: If we merge this in before the other PR it's fine, then it will be overwritten by the following PR.

PR #645, opened 2023-10-13
Fix outline button style on fossil and sandstorm variations

MaggieCabrera:
**Description**

Alternative to https://github.com/WordPress/twentytwentyfour/pull/643

**Screenshots**

**Testing Instructions**

**Contributors**

PR #646, opened 2023-10-13
Remove button border and fix outline variation padding inconsistencies

### LINKED ISSUE #613: Remove button block border
We should remove the border applied to buttons (or make it transparent perhaps). Otherwise when you change the background color, the border is still applied and there's no clear way to remove it.

CleanShot 2023-10-11 at 11 31 27

richtabor:
**Description**
An alternative to #645 and #643, which closes #613. Makes the buttons the intended size across variations, while also removing the border (which made it harder to change colors). 

This fix is applied to default  and fossil variations for now; the others do not have varying button styles (yet) so the fix in default is inherited there. 

**Screenshots**
CleanShot 2023-10-13 at 15 07 01

CleanShot 2023-10-13 at 15 14 20

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -300,23 +300,17 @@
 			"core/button": {
 				"variations": {
 					"outline": {
-						"border": {
-							"bottom": {
-								"style": "solid",
-								"width": "1px"
-							},
-							"left": {
-								"style": "solid",
-								"width": "1px"
-							},
-							"right": {
-								"style": "solid",
-								"width": "1px"
-							},
-							"top": {
-								"style": "solid",
-								"width": "1px"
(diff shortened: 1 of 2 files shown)


theme.json
```diff
@@ -811,23 +804,7 @@
 				},
 				"border": {
 					"radius": ".33rem",
-					"color": "var(--wp--preset--color--contrast)",
-					"bottom": {
-						"style": "solid",
-						"width": "1px"
-					},
```
MaggieCabrera, line 818: this makes a lot of sense, the element doesn't have an outlined variation

MaggieCabrera (approved):
Thank you!

PR #649, opened 2023-10-13
Tweak styles

richtabor:
Adds more gradients/duotones, and makes ice more unique. 

Props @beafialho.

CleanShot 2023-10-13 at 17 27 53
CleanShot 2023-10-13 at 17 28 03
CleanShot 2023-10-13 at 17 28 11
CleanShot 2023-10-13 at 17 28 21


styles/ice.json
```diff
@@ -3,56 +3,221 @@
...
+					},
+					"name": "Extra Extra Large",
+					"size": "3.5rem",
+					"slug": "xx-large"
+				}
 			]
+		},
+		"layout": {
+			"contentSize": "620px",
+			"wideSize": "99vw"
```
MaggieCabrera: this makes wide width content virtually full width. I suppose that's the intent, just pointing it out

MaggieCabrera: The home looks a little weird on really big screens

MaggieCabrera: Screenshot 2023-10-16 at 12 05 53

MaggieCabrera (approved):
my comments are not blockers, generally these look good to me

styles/rust.json
```diff
@@ -47,49 +57,6 @@
 		"layout": {
 			"contentSize": "620px",
 			"wideSize": "90vw"
```
luminuu: As Maggie mentioned above for the Ice style variation, I noticed the same with Rust too - however it was already here before. I find this a bit unlucky, as on larger screens and especially curved monitors with a wide width, these layouts would look pretty much out of place.

richtabor: I'll remove it on rust too.

richtabor:
Updated the styles to this: 

https://github.com/WordPress/twentytwentyfour/assets/1813435/8cf0b05a-96d4-4183-8cf6-54543f3afb8c

PR #667, opened 2023-10-16
Adding sans-serif and serif system fonts

### LINKED ISSUE #572: Consider adding a system/web safe font as an alternative to default fonts

luminuu:
**Description**

This PR adds sans-serif and serif system fonts to the typography section. It has to be added to every style variation that changes the typography setup from the main `theme.json`.

A system font stack is a lightweight alternative to the provided fonts that the theme is designed with.

Alternative to #573, #590, #591.

**Screenshots**

CleanShot 2023-10-16 at 18 37 02@2x

**Testing Instructions**

1. Check out PR
2. Go to site editor
3. Open any template
4. Select some text
5. Check typography / font family settings

Alternative:
1. Go to global styles
2. Open Typography section
3. See the fonts listed as in the screenshot above

### DIFF
--- a/styles/ice.json
+++ b/styles/ice.json
@@ -139,9 +139,14 @@
 					"slug": "body"
 				},
 				{
+					"fontFamily": "-apple-system, BlinkMacSystemFont, avenir next, avenir, segoe ui, helvetica neue, helvetica, Cantarell, Ubuntu, roboto, noto, arial, sans-serif",
+					"name": "System Sans-serif",
+					"slug": "system-sans-serif"
+				},
+				{
 					"fontFamily": "Iowan Old Style, Apple Garamond, Baskerville, Times New Roman, Droid Serif, Times, Source Serif Pro, serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol",
 					"name": "System Serif",
-					"slug": "system-Serif"
+					"slug": "system-serif"
 				}
 			],
 			"fontSizes": [
--- a/styles/ice.json
+++ b/styles/ice.json
@@ -137,6 +137,11 @@
 					"fontFamily": "\"Jost\", sans-serif",
 					"name": "Jost",
 					"slug": "body"
+				},
+				{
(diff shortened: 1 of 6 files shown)


luminuu:
@richtabor I just merged in the latest updates for the style variations, please have a look!

richtabor:
@luminuu I had replaced sandstorm with "Ember" in #649. @beafialho, did you want to add it back?

styles/ice.json
```diff
@@ -137,6 +137,57 @@
...
+					"name": "System Sans-serif",
+					"slug": "system-sans-serif"
+				},
+				{
+					"fontFamily": "Iowan Old Style, Apple Garamond, Baskerville, Times New Roman, Droid Serif, Times, Source Serif Pro, serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol",
+					"name": "System Serif",
+					"slug": "system-Serif"
+				}
+			],
+			"fontSizes": [
```
richtabor, line 152: fontSizes is added twice here; let's remove the addition.

styles/sandstorm.json
richtabor, line 1: I think we should omit sandstorm; as it was removed in #649 (replaced by "Ember").

richtabor (dismissed):

MaggieCabrera (approved):
Awesome! love this

PR #672, opened 2023-10-16
Use consistent pattern titles, slugs and file names

### LINKED ISSUE #548: Pattern file names out of sync with pattern slugs
**Description**
Several of the pattern files have confusing file names that are out of sync with the actual pattern slugs. While that's not a strict requirement, keeping the file names in sync with the pattern slugs makes working with them more straightforward.

For example:
* https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/template-home-business.php uses `twentytwentyfour/template-home`. So the file should probably be renamed to `template-home`?
* All the `page-*` patterns (e.g. [this one](https://github.com/WordPress/twentytwentyfour/blob/trunk/patterns/page-01-business-home.php)) use a different slug as well. Why do they include the `page-0x` prefixes in the file name?

There are some higher level concerns regarding a few of the patterns too which I shared in https://core.trac.wordpress.org/ticket/59465#comment:21, though let's focus the issue here on just the discussion on why those file names aren't in sync with the actual pattern slugs?

**Expected behavior**
Pattern file names should match the relevant pattern slug (probably minus the `twentytwentyfour/` prefix).

richtabor:
Closes https://github.com/WordPress/twentytwentyfour/issues/548. 

Should be no changes visually, other than cleaned up pattern/template/page titles, and a new page pattern for a portfolio gallery (pictured below). Essentially for when you want to add a portfolio to a site without using a blog. 

CleanShot 2023-10-16 at 15 43 57


patterns/template-search-portfolio.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Portfolio Search
+ * Title: Portfolio search template
```
MaggieCabrera, line 3: did we not say that appending "template" was redundant because the modal states clearly that we are talking about templates?

MaggieCabrera (approved):
Just a minor comment about the name for the template patterns but not a blocker. This is working as intended

patterns/template-search-portfolio.php
richtabor: I did yea, but after thinking it through and seeing it all in effect, I think it's fine appending template.

PR #679, opened 2023-10-17
Renamed Blogging search template

MaggieCabrera:
**Description**

We missed this one on our renaming efforts when we switched from using writer to blogging


huzaifaalmesbah:
Good Catch. It looks good.

huzaifaalmesbah (approved):

luminuu (approved):
Nice catch, thank you!

PR #680, opened 2023-10-18
Add trailing slash to Theme URI

swissspidy:
**Description**

All other themes have a trailing slash

**Screenshots**

**Testing Instructions**

**Contributors**

### DIFF
--- a/style.css
+++ b/style.css
@@ -1,6 +1,6 @@
 /*
 Theme Name: Twenty Twenty-Four
-Theme URI: https://wordpress.org/themes/twentytwentyfour
+Theme URI: https://wordpress.org/themes/twentytwentyfour/
 Author: the WordPress team
 Author URI: https://wordpress.org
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -1,6 +1,6 @@
 /*
 Theme Name: Twenty Twenty-Four
-Theme URI: https://wordpress.org/themes/twentytwentyfour
+Theme URI: https://wordpress.org/themes/twentytwentyfour/
 Author: the WordPress team
 Author URI: https://wordpress.org
```
huzaifaalmesbah, line 5: So, here, it may be given too. 
`Author URI: https://wordpress.org/`

huzaifaalmesbah:
I have observed that in all default themes except TT3, there is a trailing slash in the Theme URI and Author URI.

MaggieCabrera (approved):

PR #685, opened 2023-10-20
Update text-faq.php — matching <hr> border color with details block b…

neilorangepeel:
Updated the border color.

The color of the `<hr>` element was slightly off. 

The updated version matches the background with the border color of the details block using the `contrast-2` color variable. This also now matches across all style variations.

Screenshot 2023-10-20 at 22 06 44


MaggieCabrera (approved):
this change must have been lost on a merge conflict, thanks for bringing it back

PR #688, opened 2023-10-21
Added Stable tag 1.0 in readme.txt file

shail-mehta:
Currently Stable Tag is missing in Readme.txt file so I have added Stable tag 1.0

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -3,6 +3,7 @@ Contributors: wordpressdotorg
 Requires at least: 6.4
 Tested up to: 6.4
 Requires PHP: 7.0
+Stable tag: 1.0
 License: GPLv2 or later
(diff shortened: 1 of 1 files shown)


readme.txt
```diff
@@ -3,6 +3,7 @@ Contributors: wordpressdotorg
 Requires at least: 6.4
 Tested up to: 6.4
 Requires PHP: 7.0
+Stable tag: 1.0
```
MaggieCabrera, line 6: We need to also updte the style.css file and the changelog, can you please update it? Using TT3 as a reference works

shail-mehta: @MaggieCabrera 

Style.css and change log are up to date

Just Stable tag is missing so I have added

MaggieCabrera (approved):
Thank you!

PR #690, opened 2023-10-23
Minor color contrast fix

### LINKED ISSUE #687: Some Patterns texts are hard to read in maelstrom style variation
### **FAQ Pattern's Text:**

faq

### **Centered Testimonial Designation's Text:**

centered-testimonial

### **Services call to action with image on left Pattern:**

cta-service-image-left

richtabor:
**Description**

Closes #687 by switching the pattern to use an accent color as a background (which falls back to no color when no exact color slug is matched in a variation. 

**Screenshots**

CleanShot 2023-10-23 at 11 27 11

**Testing Instructions**

1. Open the site editor.
2. Switch to the "Maelstrom" style (it's blue). 
3. Add a new page.
4. Open the inserter.
5. Search for `cta-services-image-left` within the pattern inserter. 
6. See change.

**Contributors**

### DIFF
--- a/patterns/cta-services-image-left.php
+++ b/patterns/cta-services-image-left.php
@@ -7,8 +7,8 @@
  */
 ?>
 
-<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50","left":"var:preset|spacing|50","right":"var:preset|spacing|50"},"margin":{"top":"0","bottom":"0"}}},"backgroundColor":"contrast-3","layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignfull has-contrast-3-background-color has-background" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--50);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50)">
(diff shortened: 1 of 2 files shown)


MaggieCabrera:
This is not the only pattern that has contrast issues. The details block in the FAQ pattern looks even worse, and I haven´t checked all of the rest to see if there are other concerns. I will try to have a look today and see what I encounter

richtabor:
I fixed the FAQ pattern. The rest look good.

richtabor:
This highlights the need for https://github.com/WordPress/gutenberg/issues/48581 and https://github.com/WordPress/gutenberg/issues/53996 both. 

Essentially we're limited on using alternate colors throughout variations, as we can't ensure there is appropriate contrast between every variations colors—other than base and contrast. With support for both of those issues, we can have much more color variation/adaptability between variations, allowing for more creativity/variance.

MaggieCabrera:
I looked at all the patterns and they looked ok with this change and I don´t see the contrast issue in any of the rest. If we are ok with the background change in the context of the full page templates like the about one, I´m fine to get this in for RC2:
Screenshot 2023-10-23 at 15 18 47

MaggieCabrera:
> This highlights the need for [WordPress/gutenberg#48581](https://github.com/WordPress/gutenberg/issues/48581) and [WordPress/gutenberg#53996](https://github.com/WordPress/gutenberg/issues/53996) both.
> 
> Essentially we're limited on using alternate colors throughout variations, as we can't ensure there is appropriate contrast between every variations colors—other than base and contrast. With support for both of those issues, we can have much more color variation/adaptability between variations, allowing for more creativity/variance.

Absolutely, +100000000

MaggieCabrera (approved):

PR #693, opened 2023-10-24
Updated Inline PHP Documentation

shail-mehta:
Updated Inline PHP Documentation as per suggested by @aaronjorbin in https://github.com/WordPress/wordpress-develop/pull/5552


desrosj (approved):

MaggieCabrera:
Thank you both!

Issue #404, opened 2023-09-18
"Blog Home" page is the home template on a fresh install


richtabor:
Seems strange to have the "Blog Home" page be the home template on a clean install of WordPress. Related: https://github.com/WordPress/twentytwentyfour/issues/100#issuecomment-1707910960

<img width="1509" alt="CleanShot 2023-09-18 at 12 38 14" src="https://github.com/WordPress/twentytwentyfour/assets/1813435/5d753cfd-cc7b-4d6e-b3d4-8a2102aaa70b">

MaggieCabrera:
Mmmh, I thought this was the point all along. We have the query loop further down the fold. What would you expect the home page to be? @beafialho I would love to hear your input on this

beafialho:
I expected this to be the home page as well, as the business templates are "default".

richtabor:
Right, but is “Blog Home” supposed to be the site home page?

carolinan:
Are you proposing a blog home with only the posts, and a front-page.html with the other patterns?

webmandesign:
> Are you proposing a blog home with only the posts, and a front-page.html with the other patterns?

This is what I would expect from the theme. I've tested TT4 just now on an existing WP install with static front page and dedicated blog page set and I was really surprise about the behavior:
- static front page worked as expected,
- but on blog page there was the `home.html` template content displayed. I expected posts list instead.

Also, I don't quite understand why homepage pattern has to be inserted into the theme template file. Could it be just available as a whole page pattern (like it is in `patterns/page-01-business-home.php`) a user can insert into the actual static front page content instead?

I'm not sure editing the static front page content by editing the site editor template file is the best from UX point of view. As a user I don't need to know about WP theme template hierarchy and I'm surprised if my actual front page content (the content of Page post type) is overrode by `front-page.html`/`home.html` template content.

But maybe I'm talking too much from a classic theme point of view? :)

richtabor:
> Also, I don't quite understand why homepage pattern has to be inserted into the theme template file. Could it be just available as a whole page pattern (like it is in patterns/page-01-business-home.php) a user can insert into the actual static front page content instead?

No, because then headers and footers would be part of a pattern (which means you can't reuse it as an actual pattern, as you'd insert headers and footers on pages that already have headers/footers.

beafialho:
Here's a suggestion for a more traditional blogging homepage. I mocked it up using the placeholder users will see when they install the theme on a fresh new site. 

[...]

PR #706, opened 2023-10-27
Change post pattern on `home.html`, update template slugs

### LINKED ISSUE #705: Blog template `home.html` ignores main query and breaks navigation
The `home.html` template ignores the main query and by that makes it impossible to browse the posts in the blog.

The original bug report with details is available in https://core.trac.wordpress.org/ticket/59759.

Let's use this GitHub issue only for tracking in the repository. Discussion can happen on Trac as this is related to the upcoming 6.4 release.

luminuu:
**Description**

This PR fixes the issues around template names and query loops not inheriting the default loop, raised in #705 and #404. It replaces the original `posts-grid-2-col` pattern with the `posts-3-col` pattern, wrapped in a group block to align with the rest of the layout and the original h2 headline. This fixes the issues raised in #705 with ignoring the main query and breaking the navigation.

Renaming the `home.html` template file to `front-page.html` fixes the issue when a separate blog page is set, giving the users full control over the theme with the settings they make in Settings -> Reading. This was raised in #404.

In addition, I've renamed the slugs of the template and page to the full file names, so they're the same as the other templates and patterns and are easier recognizable.

I'm aware this "breaks" aka reverts the original design, but I have to agree with @felixarntz that we should aim to not break the core functionality of the main query. I'm currently not aware what the capabilities of the query loop block in terms of design are and if there can be a different design for the query, but for now we should stick to this. If users want to go back to the original design with the three posts, they can totally do so and replace the main query, but then it's their active decision.

I'd be interested in getting feedback from @MaggieCabrera, @beafialho, @richtabor and anyone in @WordPress/block-themers.

**Screenshots**
The screenshot shows my test environment with the recent changes, including how 10 posts with images and an extra sticky post would look like on the main page, with the default WordPress settings.

CleanShot 2023-10-27 at 21 33 18@2x

**Testing Instructions**

1. Check out this PR
2. Go to Appearance -> Site Editor
3. Click on Templates
4. Click on "Manage all templates" at the bottom
5. Make sure to revert any customizations if necessary
6. Check out the Front Page template
7. Open the frontend
8. Test the navigation/pagination


huzaifaalmesbah:
Thanks For PR @luminuu 
Looks good. I think need more opinions and testing.

patterns/page-home-business.php
```diff
@@ -14,5 +14,13 @@
...
 <!-- wp:pattern {"slug":"twentytwentyfour/text-alternating-images"} /-->
 <!-- wp:pattern {"slug":"twentytwentyfour/testimonial-centered"} /-->
-<!-- wp:pattern {"slug":"twentytwentyfour/posts-grid-2-col"} /-->
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50","left":"var:preset|spacing|50","right":"var:preset|spacing|50"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--50);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50)">
+	<!-- wp:heading {"align":"wide","style":{"typography":{"lineHeight":"1"},"spacing":{"margin":{"top":"0"}}},"fontSize":"x-large"} -->
+	<h2 class="wp-block-heading alignwide has-x-large-font-size" style="margin-top:0;line-height:1"><?php esc_html_e( 'Watch, Read, Listen', 'twentytwentyfour' ); ?></h2>
+	<!-- /wp:heading -->
+
+	<!-- wp:pattern {"slug":"twentytwentyfour/posts-3-col"} /-->
```
felixarntz: If this is used as part of `front-page.html` while the WordPress site is configured to display a "static front page", it would look odd, since in that case there would only be the one page, with its excerpt shown in one of the three columns.

I'm not sure how we could fix this without modifying the WordPress template hierarchy. Honestly, I find this an odd configuration of WordPress core to choose `front-page` regardless of whether that page is configured to show the blog or a static page.

Potentially we can use the `frontpage_template_hierarchy` filter to customize this for TT4 to prefer a custom template, e.g. something like:
```php
add_filter(
	'frontpage_template_hierarchy',
	function ( $templates ) {
		if ( is_home() ) {
			array_unshift( $templates, 'front-page-home.php' );
		}
		return $templates;
	}
);
```

If we did that in `functions.php`, we could have both a `front-page.html` and `front-page-home.html` template for the individual situations. `front-page-home.html` could use the content as is here, while `front-page.html` would instead replace those post columns with something similar to `page.html`, i.e. displaying the content of a single individual page.

felixarntz (commented):
Thank you @luminuu for the PR. Regarding the primary bug raised, I think this looks great.

I also am of course in favor of generally renaming `home.html` to `front-page.html` as mentioned, but there is another concern about that, see e.g. in https://github.com/WordPress/twentytwentyfour/issues/404#issuecomment-1783126578.

I added a suggestion on how we can fix that below, but it's a bit more complicated. I'm curious to hear your thoughts, but if that change would go too far, we should probably limit this PR to only fixing the `home.html` template (without renaming it to `front-page.html`) and then continue exploring a solution for the renaming separately.

patterns/page-home-business.php
carolinan: I would like to see a video of how that would work with the Site Editor. Why is it a PHP file? How do users edit it?

luminuu: I'm getting more confused the more we talk about it. I assumed this was an issue so I renamed it, and now it seems to be a bigger issue that it's renamed? I'm fine reverting this change, but I have a hard time understanding what would be the best case, as it feels like whatever we do it's not. Adding code to fix something is not my preferred idea, because then you have to further explain how things work, and it will be harder for the average user to work with TT4. Since it's a legacy product, that's not what I want to achieve.

carolinan:
The space between the posts and the CTA above the footer is very large.

With query pagination showing (Reading settings are set to show 4 posts)
Front page with query pagination)

Without query pagination showing (Reading settings are set to show 10 posts)
Front page with query pagination

carolinan:
Page 2 is working now:

Front page when navigating to page two of the blog section

carolinan:
With a static page selected in the reading settings:

Static front page)

Ideally this should only display the title (optionally) and content of the page and nothing else. There is no way right now to encourage the user to update the *template* if they changed the pages under Settings > Reading instead of the Site Editor.

carolinan:
I do not believe this is a theme problem:
When I select a static page as the front page in the reading settings, and go to Appearance > Editor, the preview is showing me a standard page template with the title, an empty featured image block, and my page content.
It is not showing me the front page template which I would expect.

MaggieCabrera:
I one hundred percent don't believe this is a theme issue. We've had the front-page vs blog template debate since the dawn of block themes, and we are still having it with no solution in sight. These are our options and our concerns:

home.html

- Blog is meant to contain blog posts so the template we get is not what users would expect
- If a user changes to display a static page, the "Blog" template is still not acting as a blog
- Pagination on the query loop of home doesn't work because the query loop doesn't inherit

front-page.html

- When a theme uses a front-page.html template, the reading settings no longer work – it doesn’t matter what you put in your reading settings, the home page will always show the content of the front-page.html template.
- If users update their home page content in front-page.html and then switch theme, they lose their content.

---

The concerns with the query loop on our current iteration of blog are valid **because** we are using the blog template. But the page is very clearly not a blog. And the 3 featured posts are NOT a query loop in the classic sense of the word. They are not meant to show more than 3 posts, they just feature the last 3 posts. We are using the query block because it's a tool in our arsenal, it's never meant to care about the reading settings because this is a static page and I think that is OK.

Not every WordPress theme needs to feature a home page that is meant for blogging, and this is a fundamental pillar this theme is built on, we are seeing a business site (though we provide an alternative blogging template and another for portfolios that are perfect replacements for the default one and that don't suffer from this issue). The fact that choosing to build a theme that deviates from the blogging paradigm and that it keeps encountering this kind of issues and concerns is pretty telling and something we need to collectively figure out, regardless of the end decision we make for TT4.

What would happen if we removed the query block pattern from the blog altogether? Is that not a valid template for a theme to use as home page? (assuming we would rename it to front-page, since it really isn't a blog) Would that not be a valid design decision for a default theme? I think that is a little concerning.

I'm not sure about the changes being made to the pattern in this PR, I would like the designer's opinion on this, particularly at this point in the release cycle.

/cc @richtabor @jasmussen @scruffian @mikachan @jffng @beafialho

MaggieCabrera:
My proposed [solution](https://github.com/WordPress/twentytwentyfour/pull/707)

beafialho:
Here's a suggestion to adapt the blog section on the business template. It looks more refined this way, I think.

blog

**On mobile:**
blog_mobile

MaggieCabrera:
@luminuu I'm going to work on this, I closed my other PR, and I will add this new pattern to yours if that's ok? I'm full on this this morning

patterns/posts-list.php
```diff
@@ -0,0 +1,62 @@
+<?php
+/**
+ * Title: List of posts
```
MaggieCabrera: We are introducing a string here so we might need to make this not be a pattern and update on a potential point release

MaggieCabrera: I removed it

MaggieCabrera:
I implemented a new query loop that inherits following Bea's design. I didn't make it a pattern because that introduces new translatable strings. We can edit this on a point release to include them as a pattern for reusing, but that is not a blocker. Does this sound like a good solution to everyone?

Screenshot 2023-10-30 at 12-46-17 wp-stable

luminuu:
That's looking really nice, thank you! I'm going to test this later today. 

Would love it @felixarntz is able to give feedback on the question I had at the requested changes.

patterns/page-home-business.php
MaggieCabrera: I think this shouldn't be fixed in the theme

richtabor: > we should probably limit this PR to only fixing the home.html template (without renaming it to front-page.html) and then continue exploring a solution for the renaming separately.


I'm good with this as well; to keep the scope/discussion focused.

felixarntz: I agree with you all that fixing it in the theme would be merely a workaround for it being fixed in core, so I'm totally supportive of not doing that. We just won't be able to use this template as `front-page.html` until that part _is_ fixed in core. I'm happy to open a Trac ticket for that part, so that at least in the future it could be reconsidered.

@luminuu 
> I'm getting more confused the more we talk about it. I assumed this was an issue so I renamed it, and now it seems to be a bigger issue that it's renamed?

It's an issue that the blog template only shows the blog posts buried far down on the page, but that's a separate and _minor_ issue compared to what I flagged in #705 / https://core.trac.wordpress.org/ticket/59759.

Given the above limitations of core, I think we should for now stick to only fixing the `home.html` template, _without_ renaming it. In other words, that _won't_ address #404, but it _will_ address #705 / https://core.trac.wordpress.org/ticket/59759.

richtabor:
Summarizing the changes: 

# Current (trunk)

### Out of the box WordPress
No changes, as homepage display set to "latest posts":

CleanShot 2023-10-30 at 10 46 27

### Homepage display to "Static page" (using sample page)

The "home" changes to use the post content of the selected page (in this example, the default "Sample Page"). 

CleanShot 2023-10-30 at 10 48 54

The page that was the homepage is now the "Blog Home": 

CleanShot 2023-10-30 at 10 49 26




# Proposed: 

### Out-of-the-box WordPress
No changes, as homepage display set to "latest posts":

CleanShot 2023-10-30 at 10 31 00

### Homepage display to "Static page"
Create a "Blog" page, and assign the existing "Sample" page as the home page in homepage display: 

CleanShot 2023-10-30 at 10 33 01
CleanShot 2023-10-30 at 10 33 21

The "Blog" page: 

CleanShot 2023-10-30 at 10 38 00

richtabor:
There's an opportunity to reduce the complexity here in core, but I do lean towards the proposal here. The out-of-the-box WordPress settings work better, and are less changes when switching the reading settings.

patterns/template-home-business.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * Title: Business home template
- * Slug: twentytwentyfour/template-home
+ * Slug: twentytwentyfour/template-home-business
  * Template Types: front-page, index, home
```
richtabor: Let's remove `index` here, so that the home business template is not suggested as a recommended index template (it's not suited well as an index template).

patterns/page-home-business.php
```diff
@@ -14,5 +14,61 @@
 <!-- wp:pattern {"slug":"twentytwentyfour/text-feature-grid-3-col"} /-->
 <!-- wp:pattern {"slug":"twentytwentyfour/text-alternating-images"} /-->
 <!-- wp:pattern {"slug":"twentytwentyfour/testimonial-centered"} /-->
-<!-- wp:pattern {"slug":"twentytwentyfour/posts-grid-2-col"} /-->
+
```
richtabor: We should still reference this new pattern properly as a pattern; to set the standard for theme creators looking at this in the future. We're not adding a changed i18n string; it's the same string as in the posts-grid-2-col pattern.

patterns/template-home-business.php
richtabor: CleanShot 2023-10-30 at 11 08 00

MaggieCabrera (approved):
This looks good to me, let's try and find a solution in core so we can bring more variety to default themes front pages

richtabor (dismissed):

luminuu:
I just have reverted the name change from `front-page.html` back to the original `home.html`, as @felixarntz pointed out in the comment above.

patterns/page-home-business.php
MaggieCabrera: Then we lose the old pattern, I was planning on making a PR to make it a pattern if we have a point release, when we don't need to worry about the strings, but we could also do this. I'm not too bothered since the other pattern was only used in this template

felixarntz:
@MaggieCabrera 
> I one hundred percent don't believe this is a theme issue. We've had the front-page vs blog template debate since the dawn of block themes, and we are still having it with no solution in sight.

I agree. FWIW I hadn't heard about this debate before, and only recently noticed that concern myself. But I think we're mixing up two different topics / issues here. The _renaming_ of `home.html` to `front-page.html` cannot reasonably done at the moment because of a limitation in WordPress core. But that's not what #705 is about (also see https://github.com/WordPress/twentytwentyfour/pull/706#discussion_r1376366687).

> The concerns with the query loop on our current iteration of blog are valid **because** we are using the blog template. But the page is very clearly not a blog. And the 3 featured posts are NOT a query loop in the classic sense of the word. They are not meant to show more than 3 posts, they just feature the last 3 posts. We are using the query block because it's a tool in our arsenal, it's never meant to care about the reading settings because this is a static page and I think that is OK.

I understand it's meant that way, but as long as it's in the `home.html` template, it breaks the expectations. However, per the above, it's also questionable to rename it `front-page.html` in the current situation because that template may _also_ show the blog - or a static page.

> Not every WordPress theme needs to feature a home page that is meant for blogging

This is the only statement in your comment that I fundamentally disagree with. I get that not every website is for blogging, and not every theme needs to focus on blogging. But every theme, especially the default theme, needs to support the capabilities and features that WordPress offers by default.

I agree there are several limitations in WordPress core that will need to be addressed to improve the situation for the future. But we cannot just break expected WordPress behavior. As long as WordPress has the setting that the home page shows the latest posts (with that being the default), we'll need to support that in a theme, especially the WordPress default theme. And again, this is a core problem more than a theme problem, but we shouldn't ignore that in the theme.

Overall, also looking at your other PR, I _think_ other than the one statement we're on the same page regarding that there are broader issues with how these expectations work in WordPress as a whole.

For me, regardless of what we expect a home page to show, there is one technical limitation of WordPress core that prevents us from achieving what we want to achieve with TT4:
* We want to ship with a static business focused home page. This makes sense to go into a `front-page.html` template, since that is the template intended for the home page.
* However, we also need to support WordPress's default configuration of showing the blog of latest posts on the home page, which may or may not be configured in that way.
* Unfortunately, WordPress core doesn't differentiate between these two configurations in deciding which template to load - it will always pick `front-page.html`.
* This used to work fine with classic themes and their PHP templates, as those themes would just have something like `if ( is_home() ) { /* Show blog posts. */ } else { /* Show static page content. */ }`. But with the block theme paradigm, that expectation of WordPress core is no longer reasonable.
* If core supported picking a separate template depending on the configuration of the home page setting, I think we would be able to address the concerns in a way that satisfy your expectations regarding what TT4 should be focused on, while also satisfying the feature set that WordPress core expects.
* That said, the `home.html` template is intended for the blog, so it _needs to_ support the capability of showing the blog posts, including navigation etc.

Obviously, we won't be able to address this core limitation for 6.4, so we'll have to find another way to work within these constraints and get to some trade-off solution. I am not sure I understand your current idea correctly, so it would be great if you could clarify that one more time. What do you propose?

patterns/page-home-business.php
richtabor: I think it's fine to leave the old pattern; we just don't use it in a template but it would still be available to insert wherever?

patterns/template-home-business.php
richtabor: I've done this in a separate PR here: https://github.com/WordPress/twentytwentyfour/pull/709

MaggieCabrera:
> Obviously, we won't be able to address this core limitation for 6.4, so we'll have to find another way to work within these constraints and get to some trade-off solution. I am not sure I understand your current idea correctly, so it would be great if you could clarify that one more time. What do you propose?

I closed that PR in favor of this one after talking to Bea and her proposing the new pattern that's included in this PR. I think front-page should be able to provide a static page regardless of query block being present at all but like you said, this is not something achievable right now. And I think that is something we should absolutely fix upstream, so this doesn't happen for future default themes and they can decide if the front page is or isn't focused on blogging.

I like that this solution is contained and it doesn't alter the main concept for the theme. Thank you all for keeping an open discussion.

richtabor:
> We want to ship with a static business focused home page. This makes sense to go into a front-page.html template, since that is the template intended for the home page.
However, we also need to support WordPress's default configuration of showing the blog of latest posts on the home page, which may or may not be configured in that way.

Doesn't this proposal resolve both of these? Where you see the latest posts on the home page, while also having a home page that is not solely a blog?

richtabor:
> I just have reverted the name change from `front-page.html` back to the original `home.html`, as @felixarntz pointed out in the comment above.

Based on the flows detailed in  https://github.com/WordPress/twentytwentyfour/pull/706#issuecomment-1785404020, the proposed PR (with the name change) seems to work better. Would appreciate feedback on the before/after, and other considerations. 

TLDR, if we use front-page.html, the home page stays the home page, when you change reading settings. You can still do whatever you'd like template editing wise, but you don't loose the homepage as you know it (as the setting reverts to using post content).

patterns/template-home-business.php
luminuu: Merged that one

patterns/page-home-business.php
felixarntz: Yeah, IMO it would be fine to keep the old pattern.

On another note, do we want to move the new markup below into a new pattern so that it is reusable, or should it remain only as part of this greater `page-home-business.php` pattern?

felixarntz (approved):
In its current state, the PR addresses the problem outlined in #705, so this LGTM.

There's an outstanding conversation to address in https://github.com/WordPress/twentytwentyfour/pull/706#discussion_r1376380759, though I'll leave that to y'all to decide on. I left my feedback there for consideration.

patterns/page-home-business.php
MaggieCabrera: I don't understand, if we create a new pattern, the field `Title:` is a translatable string. Are you suggesting that the new pattern reuses the same title that the old pattern uses?

luminuu:
I'm quickly summarizing the difference between the behaviour of `home.html` and `front-page.html`

### If "Your homepage displays" is set to "Your latest posts"
Both `home.html` and `front-page.html` show as homepage on the website.

### `home.html` and static pages set for both Home and Blog
- Home: showing the selected page
- Blog: showing current `page-home-business.php` template and blog posts show the latest posts

### `front-page.html` and static pages set for both Home and Blog
- Home: showing current `page-home-business.php` template and blog posts there show only the set page, not all blog posts
- Blog: showing the `index.html` template

Based on what's happening with the query loop when renaming to `front-page.html` I'd leave it as is with `home.html`. Thoughts?

MaggieCabrera:
I think with the changes to the query block, leaving this as home sounds good, then the user doesn't lose the settings from the sidebar to change the number of posts, they don't need to fish for those options.

felixarntz:
@richtabor 
> > We want to ship with a static business focused home page. This makes sense to go into a front-page.html template, since that is the template intended for the home page.
> > However, we also need to support WordPress's default configuration of showing the blog of latest posts on the home page, which may or may not be configured in that way.
> 
> Doesn't this proposal resolve both of these? Where you see the latest posts on the home page, while also having a home page that is not solely a blog?

The PR in its current state does address this indeed. The remaining issues are:
* When you configure WordPress to show a static front page, it won't show the current business focused template anymore - which is sort of expected, but a bit strange, given that the business template looks more like a static page than it is focused on the blog posts.
* When you configure WordPress to have a dedicated blog page (that is _not_ the home page), it'll still show this business template. That template makes sense for a home page, but not for a page expected to focus on the blog.

That said, those two problems cannot reasonably be fixed by TT4 at this point, as they are limited by the current problematic WordPress core behavior. So I don't think we should get hung up on them for this PR. I've just opened a Trac ticket https://core.trac.wordpress.org/ticket/59767 for us to be able to at least address this limitation in the (hopefully near) future.

felixarntz:
@luminuu I agree this would be best left as is, in `home.html`, given the core limitations.

> ### If "Your homepage displays" is set to "Your latest posts"
> Both `home.html` and `front-page.html` show as homepage on the website.

Maybe worth adding: Unfortunately, WordPress will choose `front-page.html` over `home.html` in this situation. If that wasn't the case, we would have been able to consider a solution where TT4 provides both `front-page.html` _and_ `home.html`: We could have renamed the business home template to `front-page.html` and replicated `index.html` as `home.html`, so that the business template would only be used when WordPress is configured to show a static front page. But that's not possible, and even then it probably wouldn't be the right solution. 🙃 

So I hope we can address this with more flexible template selection via https://core.trac.wordpress.org/ticket/59767 in the future.

patterns/page-home-business.php
felixarntz: Apologies, maybe I didn't understand the discussion fully. Is the concern about introducing new translation strings? If so, I personally don't think we should have that limit ourselves from making this a reusable pattern, but I'm aware of the late translation string changes in the release cycle not being preferable. But maybe this is an exception worth making?

Let me just say: If introducing a new translation string for the new pattern title isn't a concern, IMO it should be a new pattern, including a new title. If we want to avoid a new translation string, then I guess it's okay to keep this as is.

MaggieCabrera: I honestly don't know how much of a leeway we have when introducing new strings after freeze! but that is definitely the only thing stopping us from making a pattern here. If you think this is ok to do, then I say go for it, let's add a pattern

felixarntz: I would be in support a new pattern here. But your call of course. If someone complains about it later, all I can say is I supported the decision, but there may still be backlash at us for it 🙃

luminuu:
Quick heads up I've renamed the title of this PR as the renaming has been reverted and will most likely not make it into the merge.

MaggieCabrera:
I made the query block part into a pattern as discussed and I'm bringing this in. Thank you all for your patience getting this fixed

MaggieCabrera:
Core PR [here](https://github.com/WordPress/wordpress-develop/pull/5593)

PR #707, opened 2023-10-30
Add a blog template, rename blog to front-page

MaggieCabrera:
**Description**

This is an alternative to https://github.com/WordPress/twentytwentyfour/pull/706

I state my concerns [here](https://github.com/WordPress/twentytwentyfour/pull/706#issuecomment-1784710548). I opt for this solution that maintains the spirit of the theme while making the home template a true blogging one, even if by default the main page is a static one, which is what this theme intends to do. 

The user can still swap the front-page template to use one of the other 2 patterns that do contain a query block that inherits the query from the page they want that. 

If a user wants to create a page and assign blog posts to it, it will use the new home.html template, that has the blogging template on it.

### DIFF
--- a/templates/front-page.html
+++ b/templates/front-page.html
@@ -1,0 +1,1 @@
+<!-- wp:pattern {"slug":"twentytwentyfour/template-home"} /-->
--- a/templates/home.html
+++ b/templates/home.html
@@ -1,1 +1,1 @@
-<!-- wp:pattern {"slug":"twentytwentyfour/template-home"} /-->
+<!-- wp:pattern {"slug":"twentytwentyfour/template-home-blogging"} /-->


huzaifaalmesbah:
I notice, below the post thumbnail appears the title and excerpt
Screenshot 2023-10-30 at 5 02 10 PM

MaggieCabrera:
After talking with Bea, she is going to make a design that fits more the solution on #706. @huzaifaalmesbah can you confirm is this happens when using this pattern on a page using trunk? Cause that would be a bug

felixarntz (commented):
@MaggieCabrera If I understand your intent here correctly, I _think_ I'd be onboard with that solution.

I left a much more detailed response in https://github.com/WordPress/twentytwentyfour/pull/706#issuecomment-1785447074, but I think the approach in this PR would address the following issues:
* Sites with a dedicated blog page would have the blog posts and navigation shown as expected, which I think is the only fundamental functional limitation from a WordPress core perspective, and the reason I opened #705.
* The front page would still work well for the expectation that a static front page is shown.

It wouldn't address the situation where the home page is configured to show the blog with the latest posts. But per my reply on the other PR, I don't think we can address that right now due to the core limitation. So I think this approach here would be good as a trade-off.

PR #708, opened 2023-10-30
Make Home template pattern inherit the query

MaggieCabrera:
**Description**

the query block on this pattern was set to `inherit: false`, since this pattern is meant to be used for blogging, it should be set to true instead.


MaggieCabrera:
@luminuu

richtabor (approved):
Works as expected: 

CleanShot 2023-10-30 at 10 12 50

PR #709, opened 2023-10-30
Remove index as from unrelated templates

richtabor:
**Description**
Better surface the right templates in the right areas, instead of a wholesale approach. This reduces the confusion of switching between blog templates on the "Index" template. Surfaced while reviewing #706. 

**Screenshots**

Before: 
CleanShot 2023-10-30 at 11 15 39

After: 
CleanShot 2023-10-30 at 11 14 49

**Testing Instructions**

1. Activate the theme.
2. Go to the index template.
3. Open the Template inspector. 
4. Click on the vertical ellipsis. 
5. Click on "Replace template". 
6. See changes.

### DIFF
--- a/patterns/template-home-business.php
+++ b/patterns/template-home-business.php
@@ -2,7 +2,7 @@
 /**
  * Title: Business home template
  * Slug: twentytwentyfour/template-home
- * Template Types: front-page, index, home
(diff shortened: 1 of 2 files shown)


luminuu (approved):
LGTM

PR #719, opened 2023-11-03
Separate outline css to its own function

MaggieCabrera:
**Description**

While presenting the theme yesterday during developer hours I noticed that we are putting on the same function the registering of block styles and the enqueueing of the stylesheet for the outline block style. While the end result is similar in nature, I think it can be confusing for users that want to learn from TT4, so they are best separated into two different functions.


huzaifaalmesbah:
I agree with that looks good. It will be helpful to understand.

luminuu (approved):
LGTM

PR #720, opened 2023-11-03
Fix height of newsletter landing pattern

MaggieCabrera:
**Description**

We were using vw units instead of vh for the min-height on this pattern, this fixes it.

Before:

Screenshot 2023-11-03 at 16 42 24

After:

Screenshot 2023-11-03 at 16 42 32

### DIFF
--- a/patterns/page-newsletter-landing.php
+++ b/patterns/page-newsletter-landing.php
@@ -10,8 +10,8 @@
  */
 ?>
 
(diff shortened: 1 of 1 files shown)


huzaifaalmesbah (commented):
Maybe I've seen some hard code this pattern. Can we use preset for this?

MaggieCabrera:
> Maybe I've seen some hard code this pattern. Can we use preset for this?

I don't think so, we haven't used this value in many places, so we don't have a preset for this

PR #722, opened 2023-11-06
Resolve border colors in variations

### LINKED ISSUE #686: Separator Color Issue in ice , rust & maelstrom Style Variations
Separator Color is not visible properly in ice , rust & maelstrom Style Variations.

## **1. Ice Style Variation:** 

**Pricing Pattern's Separator is not Showing properly :**

pricing

**Sidebar Pattern's Separator is not Showing properly :**

sidebar

## **2. Rust Style Variation:** 

**FAQ Pattern's Separator is not showing properly:**

rust-faq

## **3. Maelstrom Style Variation:** 

**FAQ Pattern's Separator is not showing properly:**

maelstrom-faq

richtabor:
**Description**

Closes #686 by using base and contrast only in patterns, so that colors are persistent across variations. This is another limitation that color sets and https://github.com/WordPress/gutenberg/issues/53996 would reduce. 

Also removes the border applied by the style; instead the pattern adds the borders to the details block. This way, you can style the details block however you'd like without having to figure out where the border came from. The arrow block style is direct in that it provides an arrow (in chrome). 

**Screenshots**
CleanShot 2023-11-05 at 20 25 59

CleanShot 2023-11-05 at 20 26 16

**Testing Instructions**
1. Activate the theme. 
 2. Add a page.
 3. Search for the FAQ pattern in the inserter. 
 4. Try it with different variations active.
 5. Repeat for the pricing pattern.
 6. Also check the sidebar template in the site editor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,6 @@ As a default WordPress theme, it is important to leverage the existing design sy
 |------------|-----------------------------|-----------|------|--------|----------------------------------------------------------|
 | `Base/One` | `--wp--preset--color--base` | `#f9f9f9` | Base | `base` | ![](https://placehold.co/15x15/f9f9f9/f9f9f9.png) |
 | `Base/Two` | `--wp--preset--color--base-2` | `#ffffff` | Base / Two | `base-2` | ![](https://placehold.co/15x15/ffffff/ffffff.png) |
-| `Base/Three` | `--wp--preset--color--base-3` | `#00000025` | Base / Three | `base-3` | ![](https://placehold.co/15x15/00000025/00000025.png) |
 | `Contrast/One` | `--wp--preset--color--contrast` | `#111111` | Contrast | `contrast` | ![](https://placehold.co/15x15/111111/111111.png) |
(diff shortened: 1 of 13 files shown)


richtabor:
I think this will do. The other border top in pricing are fine as-is, as they are entirely decorative.

MaggieCabrera:
Should we remove base-3 from the palette if we are not using it, then?

MaggieCabrera:
I put the separators back to pricing but using contrast 3. I think it looks better with them even if they are a little darker than the original design. 

I also replaced base 3 with contrast 3 in the single template (it had the same contrast issue on the affected theme variations) and removed the base 3 color from all variations, since the color is no longer used on templates or patterns. This way we are not confusing the users by providing a color that doesn't really contrast how it's supposed to.

luminuu (approved):
Tested, LGTM!

julija1 (commented):
Uploading joined_video_e131ed7721e443d6b8ab9532688810d5.mp4…

PR #740, opened 2024-01-06
Fix system-serif slug type-o

RavanH:
**Description**

When trying the set, for example, the global headings font family to System Serif, the global style setting will be converted to style rule `font-family: var(--wp--preset--font-family--system-Serif);` (notice the upper case S) referencing a non-existent root variable.

**Testing Instructions**

1. Go to global style settings and change the Headings font face to System Serif.
2. Save styles.
3. Verify any headings on the front end: they will be using the browser default, not the root `--wp--preset--font-family--system-serif` (notice the lower case S) value.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -247,7 +247,7 @@
 				{
 					"fontFamily": "Iowan Old Style, Apple Garamond, Baskerville, Times New Roman, Droid Serif, Times, Source Serif Pro, serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol",
 					"name": "System Serif",
-					"slug": "system-Serif"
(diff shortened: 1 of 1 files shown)


RavanH:
See issue #741

MaggieCabrera:
We will close this repo since it's gotten out of sync with core. Can you make this PR in https://github.com/WordPress/wordpress-develop/ instead? I think it's still a legitimate fix

RavanH:
Done! Track ticket https://core.trac.wordpress.org/ticket/60325 and PR for code review https://github.com/WordPress/wordpress-develop/pull/5933 :)

MaggieCabrera:
Thank you so much!

PR #742, opened 2024-01-09
Wrap Navigation block in Row within the header.html template part

tjcafferkey:
**Description**

Now that the Block Hooks API has been released as part of WordPress 6.4 3PDs may want to insert their own blocks into the header, specifically _after_ the Navigation block. This currently causes problems because the Navigation block and it's siblings are treated as flex items due to its parent being a Row block.

This PR wraps the Navigation block within a Row block of its own. By default the behaviour of the header should remain the same and introduce no regressions but this now means plugins are able to use the Block Hooks API to automatically insert their own blocks using the Navigation as an anchor block without them being treated as flex items (e.g. Mini Cart or My Account blocks)

**Screenshots**

**Testing Instructions**

1. Activate the theme on this branch
2. Ensure that the header behaves the same on various viewport sizes and no regressions have been introduced by this change.

**Contributors**

### DIFF
--- a/parts/header.html
+++ b/parts/header.html
@@ -15,7 +15,11 @@
 		</div>
 		<!-- /wp:group -->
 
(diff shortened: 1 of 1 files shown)


MaggieCabrera:
We are going to close this repo because it's gotten out of sync with core. Can you make this change in core instead and open a track ticket, please?

